PR #10, opened 2016-10-14
Tweaked styles, fixed link and other problems in several admin pages.

tnabtaf:


dannon:
@tnabtaf I merged this but modified again all of the PLACEHOLDER_INCLUDE absolute-> relative link changes in ad2c3832639ed1843c345603d19c23e4c91f1864.  Is there a reason they were changed, or was it unintentional path manipulation when trying to get the images right?

tnabtaf:
Thanks.

If the paths in the includes were updated, it means they were root-relative in the source moin (and thus would not work in GitHub. If the paths to the includes were relative, they should have stayed unchanged.

I'll see what I can do to **not** update root-relative paths in INCLUDES.

PR #13, opened 2016-10-18
Cleanup of Admin hierarchy.

tnabtaf:
I hope to never revisit this set of pages again.


dannon:
Not sure if the change is good or bad yet, but I had previously only converted (md-style only) in-repository links when valid.  So any you are finding are actually broken links that we might want to address here prior to merge.  Or not, let me know what you think!

tnabtaf:
Hi Dannon,

Just saw this.  The changes in this PR update every link they can find, in MD or HTML.  And I haven't viewed the latest changes on a web site.  However, I did manually inspect every set of changes up through "L" pages and didn't find any big problems with the (accumulated) bug fixes.

There are still some problems I know about, but I'm hoping that this set of changes fixes 99% of the bugs in the first version.

I'll poke around some, and submit bug fixes for the renamed pages as well.

PR #18, opened 2016-11-23
Update universe_wsgi.ini -> galaxy.ini

dannon:


src/Admin/Config/GenomeSpace/index.md
```diff
@@ -30,7 +30,7 @@ First, create a list of the URLs you will need:
 
 ## OpenID
 
-In your universe_wsgi.ini file set: enable_openid = True
+In your galaxy.ini file set: enable_openid = True
```
peterjc: Missing back-ticks for pretty markup

src/Admin/Config/Performance/Cluster/index.md
```diff
@@ -60,7 +60,7 @@ galaxy_user@node1%
 ```
 
 
-If your cluster nodes have Internet access (NAT is okay) and you want to run the data source tools (upload, ucsc, etc.) on the cluster (doing so is highly recommended), set `new_file_path` in `universe_wsgi.ini` (`config/galaxy.ini`) to a directory somewhere in your shared filesystem:
+If your cluster nodes have Internet access (NAT is okay) and you want to run the data source tools (upload, ucsc, etc.) on the cluster (doing so is highly recommended), set `new_file_path` in `galaxy.ini` (`config/galaxy.ini`) to a directory somewhere in your shared filesystem:
```
peterjc: Unless you want to replace it with a historical reference to the universe filename, I would remove this bit:

> (`config/galaxy.ini`)

peterjc:
Two minor comments.

Switching to markdown like this makes site-wide changes like this much much easier than with a wiki doesn't it? You can use grep and awk for a first pass - it will make some things much easier 👍

dannon:
Yep, totally agree, it's going to be *way* easier to keep this up to date.

PR #22, opened 2016-11-30
Initial import/convert of missing creole pages.  

dannon:
These will need more cleanup, but it's a start.


dannon:
@tnabtaf This should be a first pass on #22, I used a utility to convert from creole->html->markdown, so it'll be a little shaky.

PR #27, opened 2016-12-05
Only show toc when it exists, otherwise default to full-width page.

dannon:


dannon:
@tnabtaf This should make a lot of pages without toc's look nicer.

PR #29, opened 2016-12-06
Fix some links into the top menu

bebatut:
Hi,

I fixed some broken links in the top menu.

Btw, I have a question. Is it normal that some subdirectories start with a lower case letter and other with a upper case letter? 

Bérénice

### DIFF
--- a/src/config/menu.yaml
+++ b/src/config/menu.yaml
@@ -5,13 +5,13 @@
   links:
       - News:
         title: News
-        link: News
+        link: news
       - Events:
         title: Events
-        link: Events
+        link: events
       - Support:
         title: Support
-        link: Support
+        link: support
       - Biostar:
         title: Biostar
         link: https://biostar.usegalaxy.org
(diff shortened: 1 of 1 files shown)


bebatut:
I just see that `src` directories (`events`, `news` and `support`) are used a lot of times in the pages with the first letter in upper case. Maybe is it better to change the subdirectories for upper case first letter than changing all the links. What do you think? In this case, this PR does not have any more sense

dannon:
@bebatut Thanks!  Personally I'd like to standardize towards lowercase, let's see what other folks think though.  @clements?

bebatut:
@dannon I could help on that :)
I tried also to fix the issue with the images but I can not fix it now (at least locally). I will try again later when I will read a bit about Metalsmith and how is it working.

dannon:
Merging this for now to make things functional.  We should revisit standardization of directory casing, and I've made an issue for follow up at #39.

PR #32, opened 2016-12-06
Add gitter lobby to site.

dannon:
Has display issues w/ navbar, just need to bump z-index or give it a custom element.


dannon:
Merging this to kick the tires a bit with more folks -- we can always kill it if we hate it.

PR #41, opened 2016-12-07
Fix 2 links

nsoranzo:
There are some broken internal links in the Events table, but since this is in HTML I'm not sure what's the best way to fix them.


dannon:
@nsoranzo I fixed a bug we had in html href/src handling this morning, so 'src="/src/Directory/index.md" should work fine now.

nsoranzo:
Cool, thanks @dannon!

PR #46, opened 2016-12-08
add missing Galaxy image

bgruening:


dannon:
@bgruening Are you using git-lfs for this?

bgruening:
Oo, no. Ok need to read about this.

afgane:
> Ok need to read about this.

Where?

dannon:
https://git-lfs.github.com/

dannon:
It should be basically -- install that git plugin, then ```git lfs install```

It won't change anything in your usage of repositories not using LFS, and it'll just work automatically on repos using it (like this one).

afgane:
Yes, but what about the integration/use within this site - well, I guess there's nothing on that yet? I'll add some help text to the README.

dannon:
@afgane Not sure what you mean?  Want to join in https://gitter.im/galaxyproject/Lobby?

afgane:
What I mean is when a random person wants to add a picture to the site, how will they know to (a) use the LSF, and (b) be able to do it easily without having to read any of the LSF docs.

(I'm on the gitter channel but saw this PR before the conversation there shifted that way - plus, we'll want some docs outside that channel)

dannon:
Ok, I see what you mean now.  I'm still trying to figure out how I expect most people to contribute.  I think, instead of local dev, it'll probably be through the GitHub interface, which would do the LFS bits automatically.  Definitely worth documenting though, @bgruening took a stab in #47

bgruening:
@dannon if you drag and drop an image into a folder via github than it is using lfs automatically?

dannon:
I expected that it would, but we should test this I guess.

bgruening:
@dannon any idea how we can check that this is now added with lfs?

dannon:
Well.  On the command line I can tell it is, per:
```
diff --git a/src/images/GalaxyNewLogo_GalaxyProject_Trans.png b/src/images/GalaxyNewLogo_GalaxyProject_Trans.png
new file mode 100644
index 00000000..0a145b96
--- /dev/null
+++ b/src/images/GalaxyNewLogo_GalaxyProject_Trans.png
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4c2ac4b3f0ffcb855ddf38eddc2db29b731e7790f0cb20aee0ede4768165e570
+size 56876
```

But I do not know how to do that in github, it hides this from the user quite nicely, as far as I can tell :)

bgruening:
Ok, so it is working, good to know.

afgane:
@bgruening you can also do `git ls-files file/path`. The file will either be listed or the output will be empty and exit code will be 1.

bgruening:
Thanks @afgane!

Issue #2, opened 2016-08-17
Large files


dannon:
_lots_ of large files (multiple gigs worth of pdfs, ppts, etc.) were being attached to the wiki.  We need to come up with a plan for these that doesn't involve having them in the repo.

dannon:
We're using LFS.  Jenkins does not like LFS, though, so that needs to be fixed.

tnabtaf:
LFS is currently being used for images.  We did that because many of the images from the wiki could be losslessly compressed a great deal more (and we didn't think about this until after we'd copied the originals

I think there's a good argument for leaving most images in the repo (I think only a few are bigger than 1 meg) *and for loading more in*.  Images in other websites go away, causing our older pages to look bad.

However, attachments like PDFs, PowerPoints, Word Docs and so on can be enormous, and we do have a lot of them. Wherever we put these, I don't think we need versioning for them.  (If we do need versioning, then GitHub seems like a good place, even if the attachment is huge.)

We could use commercial services (know of any?).  We could also use something like https://depot.galaxyproject.org/ (@natefoo, @InitHello) .  Either way, we need to work out a method for people to upload the files.

## Do we want to continue uploading large files (anywhere) as a best practice?

Or should we link to large files using the sites they are already on, whenever possible?

I used to do this with the old wiki because then our Google custom searches would also search the attachments.  That is worth a lot, IMO.  However, I stopped doing that for things that were already on the web a couple of years ago.  It's work (mostly time) to upload them and it is a lot of storage.

*My answer to this question depends on how much pain an upload mechanism would cost us, both in terms of time and money.*

jxtx:
I'm not sure I understand the argument against LFS (or if you are even making one). It works, people can upload images using it, images don't change a lot so having history probably isn't a big deal... What am I missing?

dannon:
@jxtx I think we're definitely set on using LFS (and, well, we already are), the question is just what to do with *very* large files that we've attached to the wiki in the past, files which range upwards of 100s of megabytes.

natefoo:
Separate repo?

tnabtaf:
Looking into GitHub, it appears to limit *repositories* to 1GB each. I can't find a user/project limit.  So, in theory, as long as our attachments sum to  < 1 GB, this might be possible.  (Hmm. We might actually approach that limit after a while.)

I can see a couple of ways to implement this:

[...]

PR #47, opened 2016-12-08
Add git-lfs setup to the readme file

bgruening:


dannon:
I think for *very* large external files, per #2, we're going to probably leave it up to folks to host however they want and simply link.  But, yes, to everything else in here.

PR #48, opened 2016-12-08
Link manipulation logic tweaks. - Work in Progress

dannon:
It's faster, in any event, though there are still broken links due to, I think, wonky scoping


bgruening:
I tested this and it does not solve the problem I see on http://0.0.0.0:8080/news :(

dannon:
Yeah, I think it's the scoping.  I need to try out some raw javascript doing the same manipulations and test.

bgruening:
This works great!!!

PR #56, opened 2016-12-08
Add link to live website in README.md

lparsons:
This is also testing editing a page directly in GitHub (as mentioned in the readme).

One thing to note, it's near impossible to keep the 80 char wide formatting using GitHub editor (unless I'm missing some feature).


dannon:
@lparsons Thanks!

PR #64, opened 2016-12-08
remove SabndBox from DaveB

bgruening:
ping @davebx


davebx:
I think those reactions should make my opinion clear.

PR #75, opened 2016-12-15
Add badge for build status

erasche:
It was mentioned that it "would build shortly" during admins chat, would be nice to link in build status.


bgruening:
Nice one! Thanks @erasche!

dannon:
Great idea, thanks!

PR #80, opened 2016-12-17
Partials implementation.

dannon:
moin includes have been converted to separate files in ```/src/partials/```, and can be inserted anywhere using syntax like ```{{> events/header }}``` to use the partial at ```/src/partials/events/header.html```


partials/Admin/Config/Performance/LinkBox.html
```diff
@@ -0,0 +1,12 @@
+<div class='linkbox'>
+**[Performance](../../../../Admin/Config/Performance/ProductionServer)**<br />
+----
+[Cluster](../../../../Admin/Config/Performance/Cluster)<br />
+[Apache|](../../../../Admin/Config/ApacheProxy)<br />
+[Nginx|](../../../../Admin/Config/Performance/nginx Proxy)<br />
+[Purge H&D](../../../../Admin/Config/Performance/Purge Histories and Datasets)<br />
```
bgruening: is this ampersand correct?

dannon: Ampersands in markdown link text should be fine, are you seeing otherwise?

The links almost certainly will need updating, however.

PR #83, opened 2016-12-31
Add timing to build, cache layouts.

jxtx:
Adds metalsmith-timer to each step of the build. This can probably be
refactored for DRY, but works for now. Pass the `cache` option to
layouts which gets passed to pug. The layouts plugin does not compile
templates, this allows pug to cache the compiled version itself. For me
the pug stage of the build goes from 70s to <1s.

Autotoc is now the slowest stage at 15s. Much more usable this way.

### DIFF
--- a/build.coffee
+++ b/build.coffee
@@ -103,13 +103,6 @@ subs = (files, metalsmith, done) ->
                 files[file].contents = contents
     done()
 
-# Extend `marked.Renderer` to increase all heading levels by 1 since we reserve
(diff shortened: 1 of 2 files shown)


dannon:
Nice catch on the caching, way faster!

jxtx:
How would you feel about making this the default in the Makefile. Nice to
have some kind of progress indicator.


-- jt

dannon:
I like that idea, I can add it in a sec.

PR #89, opened 2017-01-10
fix missing link

manabuishii:
http://sqlalchemy.readthedocs.org/en/rel_0_5/dbengine.html

and

http://www.sqlalchemy.org/docs/05/dbengine.html#dbengine_establishing

are missing.


bgruening:
Thanks!

PR #95, opened 2017-01-11
Makefile enhancements

erasche:
- For non-docker builds, don't run `npm install` every time (no one cares if jenkins takes another 1.5 seconds anyway), only when package.json changes.
- Same for bower / bower.json
- Visually separate docker specific make targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,19 +22,19 @@ build: npm-deps bower ## Builds into /build, suitable for copying to webserver.
 serve: npm-deps bower ## Serve locally for viewing
 	$(COFFEE) build.coffee --serve
 
-watch: npm-deps bower ## will serve and attempt to reload files
+watch: npm-deps bower ## Serve and attempt to reload changed files
 	$(COFFEE) build.coffee --watch
 
-check: npm-deps bower ## checks for broken links
+check: npm-deps bower ## Check for broken links
 	$(COFFEE) build.coffee --check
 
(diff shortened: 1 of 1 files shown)


dannon:
Looks good, but I'm -0 on the npm/bower tweak.  It's worth it, IMO, to ensure fully installed dependencies.

jxtx:
Only matters if there is a newer version upstream and an older version installed locally right? Otherwise

dannon:
I dunno, I see both ways.  Should have been a +/-0 I guess, an RFC-comment.  Trying to anticipate what folks might do, manually tweaked a few packages with npm and as expected they don't get 'fixed'.  The 1.5 seconds is worth the process robustness, to me, but maybe not to others.

bgruening:
I guess this was a mistake.

jxtx:
Yes, sorry. I'm +1 on these changes in fact.

dannon:
Alrighty, my concerns about the deps aren't shared, it's a merge as-is.

erasche:
Deps - newer deps are installed if package.json changes. Personally prefer better build reproducibility over newer dep versions

dannon:
@erasche Right, I was saying if someone got their node_modules or bower_components into an 'unusual' state somehow.  Moving on, though.

erasche:
Edit: ah, original comment makes more sense on a reread. Ignore my last comment.

PR #106, opened 2017-02-02
improve styling of links and code

### LINKED ISSUE #101: Improve visibility of hyperlinks.
By making the hyperlinks underlined.

screenshot 2017-01-26 12 47 07

martenson:
fixes #101 
**before**

screenshot 2017-02-02 15 40 20

**after**

screenshot 2017-02-02 15 40 31

### DIFF
--- a/src/css/styles.less
+++ b/src/css/styles.less
@@ -63,6 +63,9 @@ aside {
     background: white;
     padding-top: 70px;
     a {
+        color: #2882dc;
+    }
+    a:hover {
         text-decoration: underline;
     }
     code {
--- a/src/css/styles.less
+++ b/src/css/styles.less
@@ -47,10 +47,6 @@ aside {
     z-index: 1500 !important;
 }
 
-a, a:hover {
-    text-decoration: none;
-}
-
 .navbar-brand {
     padding: 6px 15px;
 }
@@ -66,6 +62,16 @@ a, a:hover {
 #maincontainer {
     background: white;
     padding-top: 70px;
+    a {
+        text-decoration: underline;
+    }
+    code {
+        color: black;
+        background-color: #f5f5f5;
+    }
+    pre {
+        border: none;
+    }
 }
 
 #sidebar {


dannon:
I should have commented on #101, but personally I prefer the non-underlined style, similar to how the docs look at https://galaxy.readthedocs.io/en/master/.

Certainly not a -1 on my part, I just have a slight preference the other way.  Anyone else have an opinion?  Maybe @jxtx, @tnabtaf?

martenson:
My point is that relying on color for hyperlinks is a usability hurdle - especially for longer texts. Not everybody sees that well to tell those apart.

I am disappointed in bootstrap for their approach of overriding the default hyperlink (and other) styles.

dannon:
It's a good point that everyone does it, but it doesn't make it right.  The first thing I tried to do was check two best-practices developer sites (https://developers.google.com/web/fundamentals/ and https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide) and of course they both use non-underlined links for even long form content.  

What's the actual research on it?

It does *look* better, and I could actually see an argument that underlines are *not* a net usability benefit when reading large blocks of text, might even say they're a (minor) usability hindrance because of the distraction, and that's perhaps how webpages as a whole have evolved towards this style.

Would like feedback from others, too, but maybe as a compromise we use underlines on hover, like the firefox dev site?

martenson:
@dannon After a bit deeper reading on this I am ok with settling on underline on `:hover` and changing the color of hyperlinks so they stand out a bit more.

It really does seem that things have changed and almost no trend setter prefers the underlined links anymore. Although they are superior usability-wise.

martenson:
btw checkout the TOC on right at this [page](https://developers.google.com/web/fundamentals/design-and-ui/ux-basics/). The highlighting of the current section is pretty neat.

martenson:
screenshot 2017-02-03 13 42 40

nekrut:
I like it :+1:

dannon:
Links seem a bit too bright to me, but we can give it a spin and see how folks feel about it.

PR #112, opened 2017-02-05
Streamlined public servers page

nekrut:


bgruening:
@nekrut there are many more file in this PR than the public Galaxy servers. Is this intended?

dannon:
@nekrut What did you do differently on the second commit?  It looks like all LFS'ed files got touched somehow, though I haven't dug into the differences here.  Should not be merged as-is, though I can cherrypick the differences to galaxy-public-servers if you want.

dannon:
(changes will be reopened in a subsequent PR)

PR #115, opened 2017-02-07
updated public servers

nekrut:


martenson:
I feel like we are dropping a lot of useful information like support contacts (I pointed users here dozens of times for these), citations, quotas and more. I believe we need *more* information on this page about each public sever - not less.

I understand that the 3-minutes-scrolling page as it is now is not a good information structure, however that could be remedied by responsive search and/or filtering.

dannon:
@martenson I was just typing up the same thing, I agree.  I'd like a list like this that unfolds with more details for each row.  Eventually, we probably want ```src/galaxy-public-servers/**/index.md``` with an entry for each server, and this page being autogenerated with all these details.

nekrut:
I'll modify this one

PR #119, opened 2017-02-07
refreresh of Public servers

nekrut:
I gave up - this retains all info. I just beautified the top of the page

### DIFF
--- a/src/config/menu.yaml
+++ b/src/config/menu.yaml
@@ -15,15 +15,27 @@
 - Learn:
   title: Learn
   links:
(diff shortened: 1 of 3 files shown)


dannon:
@nekrut Going to merge this and fix it up after, but the 'title' of a page should be included in yaml and automatically injected into the page in a consistent way.

nekrut:
@dannon what title is in frontmatter I do not get TOC

dannon:
@nekrut Ok, let me check that out and fix it up.

PR #128, opened 2017-02-10
Include mathjax

dannon:
We probably want to revisit this to potentially make it a page-level conditional depending on how expensive our javascript stack turns out to be.  Making it conditional is, of course, more cumbersome to the user, though.


nekrut:
MathJax will be used on a very small subset of pages, so let's just include it only to relevant pages

dannon:
(we talked on IRC, will merge for now and keep an eye on how useful it'll be moving forward -- can always remove or make it conditional)

PR #138, opened 2017-02-17
Added FDA training event Mar 6

malloryfreeberg:
I guess I don't have commit privileges yet... 
I added a new event that Mo and I are doing in March.


dannon:
@malloryfreeberg I'll add you to the list now.  Please add the corresponding file at ```src/events/<whatever>/index.md```, similar to

PR #147, opened 2017-02-22
Updated the add-tool-tutorial page

malloryfreeberg:
- Updated broken links
- Unified tone and tense of tutorial
- Clarified tool definition vs configuration file
- Updated ToC formatting (I think?)


dannon:
Thanks for the changes!

Those old linkboxes and autotoc are two different things (usually), but I think they're fine to cut here.  Autotoc is for automatically generating a table of contents with in-page links, where with those old linkboxes we mainly used for links between pages (though, usually of the same section, so admin/tools to admin/develop or something).

PR #157, opened 2017-02-24
remove include of livereload script from every page

martenson:
probably remnant of `make watch`


dannon:
Yep, that's exactly what it was from.  If we can get it to work, we can re-add it (@nekrut wants it for sure), but I won't have time to fix it for a while.

PR #163, opened 2017-03-08
Added metadata tags for all news items.

tnabtaf:
Hey @dannon, can you take a look at this?

- Every news item has a title and date now.  Date is the date the item was published or last updated.
- Dropped the newsItemHeader divs from all news items.
- Left the newsItemFooter divs in place, because I want the date displayed somewhere on the page. Might want to drop those and generate that from the date tag.


dannon:
Looks good to me

Issue #130, opened 2017-02-10
add code highlighting to code snippets


martenson:

dannon:
Trial implementation in #179

PR #179, opened 2017-03-15
Basic opt-in highlighting. 

dannon:
'highlight: true' in frontmatter to engage.  If it seems like folks like it, we can make it the default (or opt-out).  It did make some preformatted blocks look weird, to me, though.


dannon:
Addresses #130, and we can continue to refine language specification/etc.

VJalili:
@dannon +1 for enabling syntax highlighting by default.

dannon:
@VJalili I don't remember if I made it opt-in for a reason or just out of an abundance of caution; I'm not opposed to having it on by default on principle, if you'd like to change it.

PR #183, opened 2017-03-16
Fix datatypes page

bgruening:


src/admin/datatypes/index.md
```diff
@@ -3,6 +3,5 @@
 Hub page for [administering](/src/admin/index.md) Galaxy data types.
 
 * [Admin/ReferenceMAFs](/src/admin/reference-mafs/index.md)
-* [Adding Completely new Datatypes](/src/admin/datatypes/Adding Complete Datatypes/index.md)
```
martenson: why was this dropped? it lives at https://galaxyproject.org/admin/datatypes/adding-complete-datatypes/

bgruening: it was not even in the old wiki: https://moin.galaxyproject.org/Admin/Datatypes and the link was broken. Will add it back.

bgruening: Ok, now?

src/admin/datatypes/composite-datatypes/index.md
```diff
@@ -35,44 +35,44 @@ class BasicComposite(Text):
...
-
-```
+${os.path.join( input1.extra_files_path, 'results.dat' )} $output1</command>
+```
+
+If a tool is aware of the file names for a datatype, then only input1.extra\_files\_path needs to be provided.
+
+There are cases when it is desireable for the composite filenames to have varying names, but be of a similar form; for an example of this see Rgenetics below.
+
+Before continuing with this section, be sure to read and understand [AddingDatatypes](Admin%2FDatatypes%2FAdding+Datatypes).
```
dannon: This and any of the other links with %2F/etc in them probably need to be updated.

dannon:
Looks good to me now!

src/admin/datatypes/index.md
martenson: @bgruening that means hub is better than the wiki! :o)

PR #193, opened 2017-03-23
simplify and shorten menu

martenson:
screenshot 2017-03-23 13 19 17

+touch few pages


martenson:
Added a commit to support full menu until 1000px wide.

martenson:
added fix to the support page

PR #202, opened 2017-03-28
Replaced ATTACHMENT_URLs with links to attachments in depot.

tnabtaf:
Still lots of links that don't work, but this change doesn't make those any worse.


dannon:
Nice!  Spot checked a bunch and they all worked for me, thanks for moving all that stuff.

PR #222, opened 2017-04-24
Toc for events news

dannon:
@martenson This should support ToCs on news and events pages when ```autotoc: true``` is set.


martenson:
splendid, thanks @dannon !

PR #223, opened 2017-04-25
Modified event layout to have image link to event URL.

tnabtaf:
@dannon, This works, but it may not be the best way to do this.  I'm trying to have the event image link to the event's URL.


dannon:
Looks good to me!

tnabtaf:
Thanks.

PR #224, opened 2017-04-25
Better explanation of SQLite's issues for prod svc

hartzell:
SQLite has excellent support for concurrency, supporting ACID transactions is one of it's best features.

The issue in a production setting is one of performance, not correctness.  In situations where writes (updates, inserts, deletes) are frequent the 'single-writer' paradigm will reduce performance.


dannon:
@hartzell Thanks!

PR #227, opened 2017-05-03
Fixes for Visualization pages

guerler:


martenson:
this drops the https://galaxyproject.org/learn/platforms/ page without redirect, which is probably good

martenson:
Thanks @guerler !

dannon:
Looks like this was merged with all the images not LFS'd.

dannon:
@guerler Check out https://github.com/galaxyproject/galaxy-hub#handling-of-images for a bit of info regarding image handling and LFS for future PRs

guerler:
Ok, Thanks will do in the future.

martenson:
@dannon what is the way to identifying that PR did not LFSify the images?

dannon:
You can check manually if you pull it, but probably more useful is this approach: via the github UI, if you click through to one of the binary files, it'll say "This file stored in Git LFS" or not.

So, in my PR which fixed it, you can see something like this:
https://github.com/galaxyproject/galaxy-hub/blob/e49d1e359af11fbe38d5884e8794dc753881016e/src/learn/visualization/trackster/custom1.png

Compared to the original here:
https://github.com/guerler/galaxy-hub/blob/6ccdbe784f6d3ec41c0a6632db52d07e2582533b/src/learn/visualization/trackster/custom1.png

dannon:
(and +1 to someone figuring out how to make this a travis check!)

PR #235, opened 2017-05-09
First pass at renaming within-wiki named anchors to lower+hyphen.

dannon:


dannon:
(May have overreached in the replacement, need to make another pass removing a few of these changes)

dannon:
I'm going to revisit this later, with a better approach.

PR #236, opened 2017-05-09
Strip NWWLs

dannon:


dannon:
@VJalili This should fix them all up in one pass.

VJalili:
Thank you @dannon. Actually I was looking for what could be the "working" link for `<<nwwl(POST)>>`. It seems, it's just `POST`--no link.

dannon:
Yep, I think that's right.  The nwwl was preventing moinmoin from making links, so this should handle it.  Thanks for taking a look and bringing this up!

VJalili:
You're welcome :)
It looks better now, as `<<nwwl...>>`'s were misleading. Thanks !

PR #237, opened 2017-05-09
Updated nginx configuration descriptions 

VJalili:
Added some missing info for SSL configuration on nginx, fixed some types, and updated text for clarity.


dannon:
@VJalili one more change that I think would bring it inline with the other naming conventions would be to move it from `nginxProxy` to `nginx-proxy`

VJalili:
@dannon good idea; I renamed the folder.

martenson:
@VJalili you renamed it using underscore as opposed to dash. 

Also renaming should follow the advice about redirects [here](https://github.com/galaxyproject/galaxy-hub#redirects) - now the links to this page are broken.

VJalili:
Thanks @martenson; please see [this PR](https://github.com/galaxyproject/galaxy-hub/pull/243)

dannon:
Good catch @martenson, and thanks for the update @VJalili.

PR #242, opened 2017-05-15
Improve collections tutorial

MoHeydarian:
Numbered figures for future reference.
Changed wording in few sections. 

Are the 'filter from file', relabel from file', and 'concatenate' tools for collections moving to Main (soon)?
In figure 12, why are there three seemingly redundant 'Auto-assign' buttons?
In figure 16, wouldn't the called peaks be merged (and not the mapped reads)?

Not fixed: 
- references to icons in figures 4a, 6d, 14,


nekrut:
merged. tx @MoHeydarian

PR #247, opened 2017-05-19
major revision of galaxy 101

nekrut:
- Updating all figures and changing overall idea a bit
- Adding drag and drop
- Will also divide into two parts: before workflow and workflow


nekrut:
Ready to go

PR #258, opened 2017-06-02
Add first PlantTribes tutorial

gregvonkuster:


bgruening:
@nekrut what have you started here! ;)
@gregvonkuster can we convince you to migrate this in a week or so to the https://github.com/galaxyproject/training-material repo? It is not finished yet, but we have high hopes! :)

gregvonkuster:
Hmm, I spoke with @blankenberg and @martenson about this before I got started and they felt it should go here.  I'm willing to place it wherever though.  ;)

martenson:
Imo it is easier to contribute to hub than to training materials - thus I recommended using hub. If there is strong urge to unify this I see no obstacle in that, whoever feels like investing time in it can do so - this is open source after all. In the meantime there is a ready tutorial, that I think is written cleanly and can be treated as a standalone page.

dannon:
@gregvonkuster This looks nice!  I'll second what @bgruening said -- this should probably be in training-materials ultimately, but it's totally fine to host it here for now.  We're working hard to make training-materials as easy to contribute to as hub currently is.

gregvonkuster:
Thanks a bunch for the merge!  Let me know when you want it moved to training-materials.

PR #263, opened 2017-06-07
nginxProxy to nginx-proxy

yvanlebras:


dannon:
@yvanlebras Thanks!

yvanlebras:
Hi @dannon . You're welcome! I just see that the modification is not applied on the hub, but maybe it's normal, we have to wait for a rebuild or something like this ? I just see another hyperlink issue and create a PR ;)

PR #270, opened 2017-06-14
documented get real user name from username and email (for submission as real user with drmaa)

bernt-matthias:
I added documentation for the changes that might be included in galaxy with PR #4096, i.e., the possibility to use the username and email for determining the real user. Plus some changes to the external scripts.


dannon:
Thanks!  Added you to the hub editors group so you can make edits like this directly moving forward.  Treat this just like a wiki.

PR #271, opened 2017-06-14
Make AllowEncodedSlashes a requirement

lparsons:
With Galaxy 17.05, encoded slashes are used for all toolshed tool URLs and thus, without the `AllowEncodedSlashes NoDecode` setting, Apache proxy will be unable to serve Galaxy instances. Also added a note that this setting is not available in the versions of Apache distributed with CentOS 6.


dannon:
Thanks!  Also added you to the hubeditors group so you can make changes like this directly, treat the contents here just like a wiki.

PR #281, opened 2017-06-30
Update index.md for SQLAlchemy documentation URL

Miachol:
The  original [SQLAlchemy documentation URL](http://sqlalchemy.readthedocs.io/en/latest/core/engines.html) is missing. I guess it need be change to  http://docs.sqlalchemy.org/en/latest/core/engines.html.


dannon:
@Miachol Thanks!

PR #282, opened 2017-07-05
Update python-drmaa and slurm-drmaa references

Petraea:
These old links no longer work, and need updating.


dannon:
@Petraea Thanks!

PR #286, opened 2017-07-17
Rnaseq delphine

nekrut:
These updates to rnaseq tutorial initiated by @Delphine-L. Still want to add tags before this is accepted, so WIP. After it is done, I will migrate this into GTN framework.


nsoranzo:
Not sure this is still relevant @nekrut @Delphine-L ?

dannon:
@nsoranzo I don't think so.  I'm closing this out but we can always revisit it.

PR #295, opened 2017-08-16
Update index.md

Ewran:
"with each other" is not super clear, I understood the "treatments (Reb1) and controls (input) correlate well with each other" is means that correlation between treatments (Reb1) and controls (input) is good. But actually not. There are good correlations between replicates.

### DIFF
--- a/src/tutorials/chip/index.md
+++ b/src/tutorials/chip/index.md
@@ -169,7 +169,7 @@ we can then feed this matrix into **NGS: DeepTools &rarr; plotCorrelation** to g
 |![](/src/tutorials/chip/corr.png)|
 |<small>**B.** Heatmap of four samples: Treatments (Rab1) and controls (Input) are well correlated among themselves.</small>|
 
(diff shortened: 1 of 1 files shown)


src/tutorials/chip/index.md
```diff
@@ -169,7 +169,7 @@ we can then feed this matrix into **NGS: DeepTools &rarr; plotCorrelation** to g
 |![](/src/tutorials/chip/corr.png)|
 |<small>**B.** Heatmap of four samples: Treatments (Rab1) and controls (Input) are well correlated among themselves.</small>|
 
-Here we can see that treatments (Reb1) and controls (input) correlate well with each other, while correlation between treatments and controls is weak. This is a good sign implying that there is some signal on our data.
+Here we can see that there are good correlations between replicates(between Reb1_R1 and Reb1_R2, and between input_R1 and input_R2), while correlations between treatments(Reb1) and controls(input) were weak. This is a good sign implying that there is some signal on our data.
```
nsoranzo: I would keep everything in the present tense, so `were` -> `are`.

Also please add a space before the `(` open parenthesis (3 times).

PR #303, opened 2017-09-03
Update index.md

Ewran:
Not from Megakaryocyte set, but from G1E set, right?


dannon:
Good catch!  If I follow the rest of this tutorial, that's correct.

PR #304, opened 2017-09-03
Update index.md

Ewran:
it was written as
"A. Running plotFingerprint on output of multiBamSummary.".
but input of "plotFingerprint" tool is BAM file, but not matrix file (output of multiBamSummary). And I tried to run plotFingerprint on filtered data (15.), and got that SES plot.
So it should be changed to
"A. Running plotFingerprint on filtered data."


dannon:
Yep, even the screenshot shows running this on the filtered data.  Thanks!

PR #306, opened 2017-09-08
Added links and fixed image paths to work in GitHub editor.

clementsdave:
Hi All,

I'm writing doc on how to update the hub and I created this account so I would have a non-privileged account to do stuff from.  So **don't give this account any special privileges!**

However, this is a real pull request that updates text on the Community page.  Please review with the normal protocol.

Thanks,

The non-privileged Dave C

### DIFF
--- a/src/community/index.md
+++ b/src/community/index.md
@@ -1,7 +1,7 @@
 ---
 title: The Galaxy Community
 ---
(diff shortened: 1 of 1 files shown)


dannon:
Thanks, looks good!

Normal protocol for me would be to add you to the editors group at this point since I recognize the name and want to encourage treating this like a wiki where folks are free to just edit (that was the plan, right?).  But since you don't want that, I won't.

PR #309, opened 2017-09-19
virtualenv is not working anymore

jjkoehorst:
I have added some modifications, but please have a closer look as I only tested it on 1 server (Ubuntu 16.04) so far. The python virtualenv from the wget is from 2011 and was causing major issues on our server. Using the virtualenv from pip seems to have worked so far. Further testing in progress.


martenson:
This documentation seems fairly outdated :/ - sorry about that.

Your fix improves things I believe.

martenson:
Thanks for the contribution @jjkoehorst !

PR #310, opened 2017-09-19
nginx docs: add section to protect the reports app

scholtalbers:
In addition add links to the relevant reports documentation.


dannon:
Thanks @scholtalbers!

PR #314, opened 2017-09-26
Move Public server list from monolithic page to directory structure

tnabtaf:
Changes the [public Galaxy server list](https://galaxyproject.org/public-galaxy-servers/) from a monolithic page to a directory based structure, where every server has its own directory and web page.

There's a lot of followup I'd like to do with this list.  I would like to get these changes in this week.


dannon:
This is definitely great stuff.  My only real concern is about the lack of a sidebar on the page, now.

tnabtaf:
A couple thoughts:

- The page is a lot more compact, thus a side bar adds a lot less than it used to.
- I did try to add a TOC (for the high level stuff), but it didn't work it.
- I think we can go as is, both now and in the long run.

dannon:
It still seems like a *really* long page to me -- the full list of individual servers, and then all the tables after, right?  Is it maybe building differently for you?

src/public-galaxy-servers/pitagora-galaxy/index.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "Pitagora-Galaxy"
+url: "http://try.pitagora-galaxy.org/galaxy/"
+server_group: "general"
+server_links: 
+  - link_url: "http://try.pitagora-galaxy.org/galaxy/"
+    link_text: "Pitagora-Galaxy Server"
+  - link_url: "http://www.pitagora-galaxy.org/about_en"
+    link_text: "Pitagora-Galaxy Project"
+  - link_url: "/src/va/pitagora-galaxy/index.md"
```
dannon: This link doesn't look like it's working when rendered, to me, but I haven't had a chance to dig yet.  I will.

tnabtaf:
Pushed a fix for the pitagora and kmer-svm links.

src/public-galaxy-servers/pitagora-galaxy/index.md
martenson: My apologies, that is not true. ⬆️ 

There is a duplication:
http://127.0.0.1:8080/public-galaxy-servers/pitagora-galaxy/
http://127.0.0.1:8080/va/pitagora-galaxy/

But the `src` prefixed link won't work. I think we want to remove the dup anyways, so I will do so.

dannon: Yeah, it was fixed in subsequent commits.  All looks good to me now.

dannon:
@tnabtaf Looks good, definite improvement.  We can continue to iterate on it, as usual, but this is a really nice upgrade.

tnabtaf:
Aha!  It doesn't work on the individual server page:

https://galaxyproject.org/public-galaxy-servers/kmer-svm/

I'll fix and commit to master.

src/public-galaxy-servers/pitagora-galaxy/index.md
martenson: It is link at the individual server page (http://127.0.0.1:8080/public-galaxy-servers/pitagora-galaxy/) and it goes to http://127.0.0.1:8080/src/va/pitagora-galaxy/index.md which is 404. Whit applies to more pages I think, as per @tnabtaf  post merge comment.

tnabtaf:
Fixed. @dannon Please take a look at commit 4f40634829a45cccbc5839cee203f0ce5fdedf5a.  The fix involved duplicating code, and rather than do that I created a JavaScript function in `layout.pug`.  It works, but there might be cleaner / better ways to do this.

Also, once this settles, I likely add calls to this new function / whatever from other layouts.

PR #316, opened 2017-09-28
Simple public servers json feed.

dannon:
Suuuuuuper minimal example of what a public servers JSON feed would look like.  Will refine more in the morning.  (xref a thread w/ @tnabtaf, @erasche, @martenson)

Instead of an exclusion list, we should pick which elements we care to include and use that.  It should also (obviously) generate a .json file and set headers correctly, instead of .html.  This is a good starting point, though.

### DIFF
--- a/build.js
+++ b/build.js
@@ -157,8 +157,8 @@ let subs = function(files, metalsmith, done) {
 
 
 let file_staging = function(files, metalsmith, done) {
-    // We want to use metadata.md (which turns into metadata.html after
-    // processing) as the topic index.
+    // We will sometimes need to stage files post rendering as other file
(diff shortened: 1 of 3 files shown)


erasche:
Why omit any information? Why not include it all?

dannon:
@erasche Yep, metalsmith internal information like file size, template layout, etc.

dannon:
Tossing an update here before I forget it -- metalsmith-headings-identifiers is mangling the contents of the stream, so we'll want to fix that before merging this.  m-h-i is also slow, and I think we can roll the functionality it provides into the marked renderer pretty easily.

dannon:
The identifiers bit is fixed with #336, after which we can merge and move forward with this.

PR #321, opened 2017-10-02
ELIXIR establishes a new official Use Case for Galaxy

bgruening:


martenson:
maybe prepend blog PRs with `[BLOG]` ? These should probably be reviewed quickly.

Issue #221, opened 2017-04-24
suggestions on how to set up authentication with LDAP/AD should point to new method first


hrhotz:
If you look for suggestions on how to handle external authetication on this page: https://galaxyproject.org/admin/config/
you first see "Apache External User Auth" and "Nginx External User Auth"  and only if you keep reading, you see the link for "Authentication with LDAP/AD" which explains the new (preferred?) method. 

Would it make sense to change the order somehow? 

Hans-Rudolf

PR #328, opened 2017-10-07
Put AD/LDAP auth link in the production section..

scholtalbers:
in the admin/config page as indicated by @hrhotz #221


dannon:
Thanks!

PR #336, opened 2017-10-25
Strip use of headingsidentifier, do our own via extension of marked r…

dannon:
…endering of titles.

@tnabtaf This trims a very expensive step out of the build and rolls it into markdown rendering.  It'll also fix our problem with those mangled quotes.

( now that I think about it, we can probably do this easily enough without the extra elements at all, if I'd just approached this thinking about the goal instead of the problem in question -- it's a step in the right direction though, and it does fix the issues )

### DIFF
--- a/build.js
+++ b/build.js
@@ -4,6 +4,7 @@ let fs = require('fs');
 let path = require('path');
 let hb_partials = require("handlebars");
 let marked = require("marked");
+let slug = require("slug");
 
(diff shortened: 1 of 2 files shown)


dannon:
Doing a little refinement, then this should be good to go.

dannon:
@tnabtaf Going to just merge this and the json feed PR; seems to be working well for me.

PR #337, opened 2017-10-30
Move Mallory to Former Members

nsoranzo:
Sorry to see you go @malloryfreeberg!


malloryfreeberg:
Happy to have been a part of Galaxy for the time that I was :)

PR #352, opened 2017-12-18
Fix a link to CyVerse in documentation

bewt85:
The old link was interpreted as relative and so pointed to a page which didn't exist.


bgruening:
Thanks @bewt85!

PR #356, opened 2018-01-03
Redirect 8 migrated admin topics to Galaxy docs.

jmchilton:
More updated content is available in Galaxy docs as well as versioned information for different Galaxy's (currently 17.09 and dev). Added to Galaxy with https://github.com/galaxyproject/galaxy/pull/5135.


nsoranzo:
Thanks @jmchilton!

PR #360, opened 2018-01-18
Fix typo in documentation for 'interval' datatype

pjbriggs:
The documentation for the `interval` datatype refers to `COMMON` as one of the columns; the correct name for this column is actually `COMMENT`.


martenson:
Thanks @pjbriggs !

PR #362, opened 2018-01-19
Update deepTools/MACS2 description

dpryan79:


bgruening:
Does this mean we will the a tutorial about ATAC-seq on training.galaxyproject.org? :)
Thanks @dpryan79!

dpryan79:
For our purposes, the difference between normal ChIP-seq and ATAC-seq is one step (`alignmentSieve`, which will be in deepTools 3). Having said that, I'd want to get Wellington or some other footprinting program in Galaxy before make a "real" ATAC-seq tutorial.

PR #371, opened 2018-01-30
Collaborate: non-GCC projects, additional pointers

chapmanb:
Dave -- this looks excellent. I added some additional color around why
we moved the collaboration to the end, and also provide pointers to
non-GCC projects. I moved that up the list a bit so folks definitely
know this is not Galaxy specific and we welcome anyone interested in any
open source bioinformatics project.


tnabtaf:
Thanks @chapmanb.  That's just what we needed.

PR #372, opened 2018-02-05
place holder for sg2018

hrhotz:
will fill with more content over the next few months


tnabtaf:
@hrhotz: Woohoo!  (Of course, I suspected as much, but it's good to see.)

Any chance I can convince you that 2018-sg would be a better directory name?  With the exception of GCC's, we've switched to putting dates first.

hrhotz:
well, I just tried to be consistent, we have already ~/events/sg2014/, and ~/events/sg2016/, plus ~/events/switzerland2012/, /events/switzerland2014, and /events/switzerland2016 as well as ~/events/germany2014/ and ~/events/germany2016/.....
Since the base of the URL changed anyway (from wiki.galaxyproject.org/Events/ to galaxyproject.org/events/) it is probably a good time to change ?  ping @bgruening

bgruening:
I'm ok with changing :) Thanks @hrhotz!

hrhotz:
excuse my ignorance with git.....is there a simple way of changing the  directory name? Or shall I just drop this request and make a new one

bgruening:
If you go to the edit button, you can change the filename, right? If you type backspace you can also change the dir name.

hrhotz:
doh!  it is just a commit on the branch itself  doh!
 @bgruening  thank you for the pointer

hrhotz:
shall I merge...in order to make the url from the logo functional?

tnabtaf:
Thanks @hrhotz. Should show up shortly.

PR #373, opened 2018-02-08
Reenable watch

dannon:
WIP, doesn't work yet (menu loading fails -- probably rewrite to be static in template)


dannon:
@afgane This should do it.

PR #400, opened 2018-05-01
added BOSC LBLT deadline

nlharris:
(you can move or reformat that added section if you want)

Also made a few tweaks to wording.


tnabtaf:
Thanks @nlharris

PR #405, opened 2018-05-10
Two Galaxy Interactive Environments

dfornika:


martenson:
There are actually many more now https://github.com/galaxyproject/galaxy/tree/dev/config/plugins/interactive_environments

martenson:
we could probably rewrite the page to be just top overview and link to https://docs.galaxyproject.org/en/master/admin/special_topics/interactive_environments.html, where the setup is described

dannon:
Going to merge this and follow up with another commit updating the tally and linking to the other config.  Thanks!

PR #408, opened 2018-05-23
Render events information in (bio)schema(s).org JSON-LD format

njall:
I've added some code to render out events in JSON-LD format according to the schema.org/Event specification. 

There are two additional fields in there called _contact_ and _eventType_. These fields aren't recognized by schema.org - but they are currently being proposed as new properties by the Bioschemas group to schema.org (See: [their recommendations to schema.org](http://bioschemas.org/specifications/Event/specification/)

[Here is a tool to test the JSON-LD formatting](https://search.google.com/structured-data/testing-tool)

### DIFF
--- a/layouts/events_index.pug
+++ b/layouts/events_index.pug
@@ -1,6 +1,7 @@
 extends layout.pug
 
 mixin renderEventRow(item)
+    +renderEventJSONLD(item)     
     tr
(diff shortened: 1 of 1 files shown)


dannon:
@njall Thanks for this, it looks great!

PR #409, opened 2018-05-25
Fixed mistake in duration format

njall:
Duration format must start with P for __period__.
See Duration ISO Standard https://en.wikipedia.org/wiki/ISO_8601#Durations


dannon:
@njall Thanks!  I didn't notice this in the validation output before.

PR #410, opened 2018-05-29
minor corrections

nlharris:
I also wanted to edit the header to say "Advance Registration ends June 1" rather than "Early Registration ends May 11" but wasn't sure where to find it.


martenson:
thanks @nlharris, I changed the registration you mentioned in https://github.com/galaxyproject/galaxy-hub/commit/e9df56eda07d49f62059652153a7e7b6d6349d11

PR #415, opened 2018-06-19
Add a training blog post 

bebatut:
Hi,

I wrote a blog post to summarize what happened for the training material over the last year.
This idea is to give an overview, so people can see what's going on. 

I would like to have your feedbacks!

Bérénice

(Spoiler alter: My talk at GCC will be quite similar)


dannon:
@bebatut Thanks!

PR #429, opened 2018-08-08
Upgrade to bootstrap 4, webpack, major build upgrades and some prettification across the board.

dannon:
This overhauls the build process pretty significantly.  Hub now uses webpack for js and style bundling/distribution, and ships bootstrap 4 with upgraded styles.  For consistency, I'm using the same styles from Galaxy, so it'll be easier to keep in sync moving forward.  I still need to revisit and cull styles we don't use at all in the hub-custom style portion, but that can wait for another PR.

To fix the "google custom search refusing to index zotero" issue, I also added a new custom zotero search, here:

image


tnabtaf:
I have some questions after poking around some.

* the style "thumbnail" does not appear to be defined, causing some of the images in `src/splash/community/index.md` to be rendered too big.
* Top menu bar doesn't convert to hamburger icon as window narrows
* Top menu bar scrolls off the page.
* I believe the class `center-block` is undefined.  Class `center` is still defined.
* Bootstrap tables now have a line above the header row, which makes things look denser than with BS 3.  Can we add a margin to the top of tables?

That's all I noticed.

dannon:
@tnabtaf Thanks for checking through it, give me a few to get these tweaks in.

dannon:
@tnabtaf Addressed all that and more.  Merging this so it doesn't go stale, with so many changes.  Happy to fix anything else we want to change as well, of course.

jxtx:
Fixes lots of things, but can we move back to having the navbar content in a container (so, same width as page content).

dannon:
Sure.  I kinda liked the full width, but if folks want to constrain it to page-width, I can put that back.

Issue #427, opened 2018-07-27
Integrate Cloud services, VMs, and Containers into Public Galaxy Servers infrastructure


tnabtaf:
This is a proposal.  Feedback is encouraged.

I'd like to update the [infrastructure we've been using](https://github.com/galaxyproject/galaxy-hub/blob/master/CONTRIBUTING.md#public-galaxy-server-metadata) for the [public Galaxy Servers list](https://galaxyproject.org/public-galaxy-servers/) to include cloud based services, virtual machines, and Docker containers.  Right now, only public servers are handled well.  There are a couple of ways to do this.  Here's a first proposal.

- Add new types like "cloud service", "VM", "container" to the supported list of types (which is currently "general", "tool-publishing" and "domain"
- Create a new entry for everything we know about.
- Render the new types as new sections on the [public Galaxy Servers page](https://galaxyproject.org/public-galaxy-servers/) 

It would be trivial to switch to listing them on a parallel page instead.

I like this idea because it much better addresses the different ways that people can run Galaxy without having to install their own.

I'd combine this work with other pending work.

Thoughts?

tnabtaf:
@dannon @martenson @erasche: ping.

erasche:
Sounds fine to me!

dannon:
@tnabtaf This sounds good to me, let me know if I can help.

tnabtaf:
Thinking out loud about how and why to do this.  

**Use Models**

Current classification scheme has `general`, `domain`, and `tool-publishing`. Those are all about the *contents or purpose* of the server.  All the servers have the same *use model* - goto a website, and start using it.  This proposal is about adding new use models like "regional cloud", "vm" and "container". 

**Many Use Models, Many Purposes**

Furthermore, a "server" can be available in multiple use modes, such as an online public server, and as a Docker container.  I can also imagine a "server" having multiple content/purpose classifications too (well, I can't imagine, but keep it in mind for the future).

**How to list that in the Hub?**

Current page shows 3 separate lists, one each for `general`, `domain`, and `tool-publishing`.   We could just add an "Availability" column that lists things like "Public sever, Container".  That works fine for the current list of servers, but what about services like PL-Grid or things that are only available as containers? 

How about:

| Server / Service | Category | Platform | The rest of the info |
| ---- | ---- | ---- | ---- |
| ARGalaxy | Domain | Public server | Immuno stuff |
[...]

PR #430, opened 2018-08-13
Use directory prototype

tnabtaf:
Hi All,

This is not ready for prime time yet, but I would like to get it on the main hub so that I can easily get feedback from community members.  The new `use` directory is not linked to from anywhere else, and the old `public-galaxy-servers` and `galaxy-services` directories are untouched in this PR.

I would just like to get this out there and solicit feedback before porting everything over and then doing the swap.


frederikcoppens:
I definitely like the pages for each server better (so the newer version for things like https://galaxyproject.org/public-galaxy-servers/abims/) 

Having badges for the platforms would be nice and take less place (and less clutter). This is probably more difficult for the scope. But it currently is already better compared to all the duplication in the original version with the links column.

What is/are the aim(s) for this list ? If it aims to help users find an appropriate instance, I'm afraid it's hard for them to do so. Is it feasible to have an interactive table where you can filter ? Can we tag instances so they can be filtered ?

You also mention the galaxy-services: these are not included in the new page, or did I miss this ?

martenson:
now available here: https://galaxyproject.org/use/

tnabtaf:
@frederikcoppens Moving conversation to issue #427.

PR #431, opened 2018-08-20
replace FR Galaxy link

erxleben:
replaced http://www.bioinf.uni-freiburg.de/Galaxy/ by https://usegalaxy.eu/freiburg


dannon:
@bgruening Of course, the intent is that it's like a wiki, we'll add anyone that wants to edit!  Added @erxleben.

PR #432, opened 2018-08-22
Add Matthias to IUC member list

nsoranzo:
See https://github.com/galaxyproject/tools-iuc/issues/1993 .

@bernt-matthias Can you check your details are correct, please?


bernt-matthias:
Jep. Everything correct.

PR #441, opened 2018-09-28
Add first "tutorial of the month" post

bebatut:
Hi,

This post will be (hopefully) the 1st one of a serie: The Tutorial of the Month.

@MoHeydarian @tnabtaf can you tell what do you think about the structure and the content?
@mblue9 can you check if your answer is correct?

Bérénice


bgruening:
I think we should mention the contributors to the training as well.

tnabtaf:
+1

PR #444, opened 2018-10-11
Added Australian stratum-1 mirror to list

nuwang:


erasche:
@nuwang could you also add it to https://github.com/galaxyproject/ansible-cvmfs?

nuwang:
@erasche Thanks for pointing me to that. Fixed in: https://github.com/galaxyproject/ansible-cvmfs/pull/8

PR #445, opened 2018-10-12
Add October "tutorial of the month"

bebatut:
I made it earlier this month to have it in the middle of the month and limit overlaps with the COTM post.

@yvanlebras I added 2 screenshots. Let me know what do you think about them

@tnabtaf, @MoHeydarian can you check if the content and structure are ok?


yvanlebras:
Thank you!

PR #448, opened 2018-10-17
Roll link-fixing into custom marked renderer (this handles yaml front…

dannon:
…matter-then-marked-rendered local links)


dannon:
@tnabtaf Check this out and see if it handles the links the way you want.

PR #449, opened 2018-10-17
 Replacing old directory pages with redirects that point to the new use directory.

dannon:
@tnabtaf -- this is just opening a standing PR from the 'redirect' branch so it's easier to see the changes at a glance -- no need to merge until it's ready.


tnabtaf:
Thanks, and thanks for the explanation.

PR #450, opened 2018-10-22
Fontawesome 4->5 upgrade

dannon:


tnabtaf:
@dannon: Please take a look at recent push.  I think it's ready to go.

dannon:
@tnabtaf Looks good to me, and I dropped the shim.  The 'far' classes said they weren't usable without pro, but it looks like they automatically map to the free fonts anyway, so we're good to go.

PR #453, opened 2018-11-02
Link to webpage with PDF in multiple languages

nsoranzo:


nsoranzo:
@tnabtaf I usually do that, I asked for Bjoern's review only because that's about a document in German for the European Galaxy Days in Freiburg and preferred to be sure.

bgruening:
Ha, perfect thanks @nsoranzo!

tnabtaf:
@nsoranzo I'll restrain my eagerness in the future.

PR #468, opened 2018-11-21
Annual site wide image compression.

tnabtaf:
Image compression was done with ImageOptim running under MacOs.  Compression was attempted on every png and jpg file in the hub. These files were compressed.

I've spot-checked a bunch of files and the compression has been successful (and appears lossless) for every one I have checked.


dannon:
@tnabtaf Looks good to me!

PR #481, opened 2019-01-16
Add tutorial of the month for 01.19

bebatut:
Hi,

With a bit of delay, here is the first tutorial of the month for 2019.
Thanks @Slugger70!

Bérénice


bgruening:
Merci!

PR #485, opened 2019-01-24
Update of the trainings for GCC2019

bebatut:
ping @bgruening 

The form is updated in the same way


tnabtaf:
@bebatut Working on this in branch.  I may throw it back at you, or just push it through.

PR #494, opened 2019-03-10
corrected data manager test command

bernt-matthias:
I don't merge this myself to add the question if there are any plans to improve &/ move the data manager docs?


dannon:
@bernt-matthias Thanks for the fix!  I don't know of any plans, though if you'd like to improve them I'm sure folks would be appreciate it :)

PR #516, opened 2019-04-25
Attempt to use only lts node from travis

dannon:
WIP, going to iterate fixing the build here.  We want to pin against the latest LTS node for stability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 sudo: required
 language: node_js
-services:
-  - docker
 node_js:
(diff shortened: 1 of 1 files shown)


dannon:
Easier than expected

dannon:
This got the travis side of the build working, and it's *much* faster without the docker part, but it's worth considering we might want to force docker usage here anyway to test the container build process.  In this instance, it was only the test travis build that failed while Jenkins kept working fine with the *real* build and deployment.

PR #517, opened 2019-04-25
Lock node docker version in Makefile

dannon:
WIP again, testing locking docker version down in the makefile.


dannon:
This is a better fix than #516, since this way we'll actually test the version of node locked in the makefile, which is what jenkins uses for the real publishing build.

PR #519, opened 2019-05-02
Update matshead layout

afgane:
@tnabtaf this is a follow up to our conversation at the team meeting a couple of weeks ago. Let me know what you think.


afgane:
The proposed layout, along with the two sub-menus that were also updated a bit:

image

martenson:
thanks @afgane !

PR #522, opened 2019-05-08
annotate platform server jurisdiction

erasche:
cc @tnabtaf @martenson, for use in public-galaxy-servers


erasche:
@dannon the json feed doesn't seem to be working? https://galaxyproject.org/use/feed.json

PR #551, opened 2019-08-13
Fixing GKE Helm installation tutorial

almahmoud:
Main changes:

-Separating Helm installation from helm initialization
-Updating to latest working Helm version on GKE, and specifying the used working versions
-Adding information about cluster context (and the fact that this can be run from any shell with the proper context not necessarily from within GKE Cloud Shell)
-Removing manual `kubectl` commands to reset `helm`
-Adding information about how to make sure a previous `helm` installation is not already present before installing a new one (to avoid the problem of confusing two different installations)


tnabtaf:
Ping @afgane @nuwang and @VJalili.  Or not.

PR #555, opened 2019-09-09
Add event for October Gallantries workshop

bebatut:


tnabtaf:
@bebatut Thanks for submitting this, and apologies for the delay.  In the future, you should be able to push updates directly to the hub, unless that makes you uncomfortable.

PR #561, opened 2019-10-29
Updated data table object path access method.

hepcat72:
This is based on feedback received (https://github.com/galaxyproject/tools-iuc/pull/2655#discussion_r340049857) after using the above suggested code based on this documentation.


dannon:
Thanks!

PR #573, opened 2020-02-11
adding (first) SIB training course

hrhotz:


hrhotz:
@tnabtaf Thank You!   Let's see how this story will develop....

PR #576, opened 2020-02-13
More than biomedical

hexylena:
So I rewrote some of the bullet points based on the [training material's]https://training.galaxyproject.org/training-material/topics/introduction/slides/introduction.html#4) intro slides to clean up the language a bit.

I removed biomedical since we have so much more and I don't want to give anyone the idea that galaxy isn't for them.

After my chat with @annefou the other month I tried to mention the community focus but I'm not sure if I've done that in a way that's communicating what we want?

@tnabtaf would definitely appreciate your feedback here :)

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -3,14 +3,13 @@ title: Galaxy Community Hub
 layout: homepage.pug
 ---
 Galaxy is an open, web-based platform for accessible, reproducible, and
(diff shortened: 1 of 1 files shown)


annefou:
@hexylena It is very good. Thank you!

PR #583, opened 2020-03-11
WIP fix search, lots of dep updates, etc.

dannon:


src/gcc/index.md
```diff
@@ -3,6 +3,7 @@ title: Galaxy Community Conferences (GCCs)
 ---
 
 <div class="text-center">
+
```
dannon: @tnabtaf Marked got more strict in its parsing and these are required now.   Unfortunately they're probably scattered around the hub, so if you see un-parsed markdown inside html, this is the fix.

src/gcc/index.md
```diff
@@ -27,21 +29,21 @@ GCC is a community driven event.  Training topics are nominated and then voted o
...
-| June 30 - July 2, 2013 | [GCC2013](/src/events/gcc2013/index.md) | Oslo, Norway | [University of Oslo](http://www.uio.no/) | 219 |
-| July 25-27, 2012 | [GCC2012](/src/events/gcc2012/index.md) | Chicago, Illinois, United States | [University of Illinois at Chicago](http://uic.edu/), [University of Illinois at Urbana-Champagne](http://illinois.edu/), [Computation Institute](http://http//www.ci.uchicago.edu/) | 202 | 
-| May 25-26, 2011 | [GCC2011](https://galaxyproject.org/gcc2011/) | Lunteren, The Netherlands | [NBIC](https://www.nbic.nl/) | 148 |
-| May 15-17, 2010 | [GDC2010](https://galaxyproject.org/dev2010/) | Cold Spring Harbor, New York, United States | [CSHL](http://cshl.edu/) | 69 |
+GCC has met annually since 2010 and the program has grown and transformed each year.  This year the  [Bioinformatics Community Conference 2020 (BCC2020)](/src/events/gcc2019/index.md) will again bring together the Galaxy and BOSC communities into a single event, featuring eight days  of training, presentations, networking, and a CollaborationFest.
+
+Dates | Event | Where | Hosts | #
+---- | ---- | ---- | ---- | ----
```
dannon: @tnabtaf Marked got more strict in its parsing and the table layout is less forgiving.  There are almost certainly more of these that we'll see (as un-parsed markdown table-ish things) and will need to 'fix'.

tnabtaf: Wow, so the leading and trailing pipes need to be dropped?  That's a terrible change.  

I can write a parser to address this.  (Just ran a site-wide updated yesterday)

src/gcc/index.md
tnabtaf: Crap, crap, crap.  

Writing a parser to catch this will be harder, as HTML can span multiple lines, and / or have markdown on the same line.  And it probably applies to all HTML?

dannon:
@tnabtaf I cherrypicked the search fix out of this and pushed it to master.  We can sort out the rest later.

tnabtaf:
Thanks @dannon 

I'll get here....

dannon:
I ported the 'search fix' out of here independently and since we have our sights set on a new platform I'm not going to bother fighting with the markdown compatibility updates here -- closing this effort out.

PR #585, opened 2020-03-19
Add documentation on Okta OIDC

selten:
This is to add documentation for https://github.com/galaxyproject/galaxy/pull/7497. It adds documentation on how to set up Okta as IdP for Galaxy.


VJalili:
@selten thanks for the docs. Could you please re-upload all the figures via git LFS?

PR #624, opened 2020-07-08
Jtech

nekrut:
Adding information about CSHL/JTech scholarship


tnabtaf:
Changes have been made via direct update.

PR #651, opened 2020-11-04
Created a Webinar events sub-page.

tnabtaf:
Created an events sub-page for webinars only.  And thus the Galaxy Webinar Series is born.


tnabtaf:
Bea, what do you think?

beatrizserrano (approved):
looks great, thanks @tnabtaf!

PR #661, opened 2020-11-18
update calendar links and utc offset

blankenberg:
No warranties on proper time zone usage is implied.


tnabtaf:
@blankenberg Whoa.  Two fixes for this came in simultaneously.  And yours lost the push/commit/merge race.  So, I am closing this PR.

blankenberg:
Bad idea to trust me about time zones anyway :) 

Not sure if the UTC offset in the text should still be updated to +1 vs +2 due to daylight savings

PR #676, opened 2020-12-31
Revamp citing-galaxy page.

tnabtaf:
Dropped emphasis on pubs from US team.  We are global and this page now reflects that better.

@blankenberg **or** @beatrizserrano. Would you mind taking a look at this?  Just want to make sure I am not missing anything obvious.  Thanks.


beatrizserrano (approved):
Looks good to me! (if I ignore my need to replace all the double spaces by single ones ;) ) 

The only part that I see redundant is the [reference to platforms descriptions in the example section](https://github.com/galaxyproject/galaxy-hub/pull/676/files#diff-35ee87d1aef13fa3e18346b178a421941244e83ca58e99039d8c0132347d2924R33).

Other than that, all good, thanks a lot @tnabtaf!

PR #687, opened 2021-01-21
Move Jobs in nav menu from own category to inside Community drop down.

### LINKED ISSUE #673: Move Jobs from top menu to be inside Community menu.
Currently the [Jobs](https://galaxyproject.org/jobs/) link is in site menu in the top bar.  It would be better under the **Community** submenu.

We also need to convert the listed jobs to be more like other lists (think [events](https://galaxyproject.org/events/)) where jobs on other sites can be easily linked to.  But that is more than a starter issue.

NickSto:
Fixes #673


tnabtaf:
Thanks you @NickSto !

PR #688, opened 2021-01-28
add GTN link for ATAC-seq nomination

hrhotz:


shiltemann:
Thanks @hrhotz! Sorry I missed that one!

PR #692, opened 2021-01-28
Cofests

tnabtaf:
@NickSto please consider this offering, which creates a new page for CoFests, and a new Page for Papercuts too.


NickSto (approved):
Looks good to me!

Only thing missing is a link to these pages from somewhere, but that can be done later.

PR #706, opened 2021-03-05
Minor tweaks: add enis and attributions to the diagram

fubar2:
Adding Enis and an expanded attribution on the diagram


tnabtaf (approved):
Thanks for catching this Ross.

PR #710, opened 2021-03-06
remove life from sciences on the community index - galaxy is agnostic…

fubar2:
… WRT the relative merits of biology vs physics

### DIFF
--- a/src/community/index.md
+++ b/src/community/index.md
@@ -1,11 +1,11 @@
 ---
 title: The Galaxy Community
 ---
-A vibrant community of contributors is the reason this project exists. The [Galaxy team](/src/galaxy-team/index.md) alone cannot possibly [maintain enough servers](/src/use/index.md), [wrap enough tools](https://toolshed.g2.bx.psu.edu/), [teach enough workshops](/src/events/index.md), implement [all needed features](https://github.com/galaxyproject), or [answer every question](/src/support/index.md) on its own. This is why community is vital to our core mission - enabling reproducible research in life sciences and beyond. Members within the Galaxy community are located all around the world and often form subgroups that differ in size (small versus large) and degree of privacy (public versus private). This page highlights these groups.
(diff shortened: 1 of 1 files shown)


fubar2:
Am I the only one that thinks this matters?
I think it's important to push the fact that Galaxy works for any data intensive sciences.

gregvonkuster:
@fubar2 No, I think that it matters a lot too!  Part of my job is to take Galaxy outside of biology.  :)

fubar2:
Excellent. Next stop world domination. But first, we must conquer the front page :)

tnabtaf:
@fubar2 and @gregvonkuster: Hmmm.  I like starting with life science because that is what pays most of the project's bills.  However, @gregvonkuster has the ability to approve this PR, and if you do that, I will pretend not to notice.  :-)

(@fubar2 I don't have the ability to add you to the organization, but I bet @gregvonkuster does.  And then you could just go around me and my old-fashioned ways!)

fubar2:
` I like starting with life science because that is what pays most of the project's bills`

Fair enough, but grants are an indicator of sucessful submissions years ago when indeed it was mostly life sciences, and no, I'd rather convince you by sound argument than work around you :)  If that doesn't work, holding my breath until I turn blue remains an option.

Given recent expansion and enormous potential, do we really need to start out by explicitly declaring that it's for life sciences? 
Is there any possible cost, grants or otherwise from stating that Galaxy is a good fit for any data intensive science. 

I worry that astrophysists or atom smashers or whatever passing by might be immediately put off reading further once they see the claim that Galaxy does life sciences. It's true but not the whole story and we are selling ourselves short. Given that it seems to work well for any data intensive science let's make sure passers by don't get the wrong idea?

tnabtaf:
And I just realized to my horror that I am arguing with @fubar2 and @gregvonkuster!  Who does that?  :-)

So, it's going in!

yvanlebras:
This matters a lot! Maybe I can add that Galaxy continue to be mainly life sciences oriented notably due to original scope and funding. And "technically" / practically one last (?) clue of this orientation is the genome / genome build mention by default notably on datasets... Is there a plan to give the possibility to make this "optional" (for example a user can choose the main use of Galaxy and say what is a "by default reference" for example genome build for omics, taxononimc+geographical+temporal coverages for ecology,... )? Just because this is something I am thinking about since I am on the Galaxy Ecology side ;)

Hope all is ok for you !

gregvonkuster:
@yvanlebras Yes, let's make displaying the ```database: ?``` on history items configurable!  I've seen several conversations about this in the past.  :)

@tnabtaf I'm not authorized to add new members either (at least I don't think I am), but that's quite ok!  :)

PR #713, opened 2021-03-10
Use homepage video as included html to demonstrate kicking out of pug…

dannon:
… into a reusable include


tnabtaf:
Disagree with YouTube playlist.  Videos can come from community as well.  They don't event have to be on YouTube.

PR #719, opened 2021-03-31
update times, zoom link for EU PaperCuts and add links to WGs

beatrizserrano:


beatrizserrano:
@tnabtaf can you check that the WGs are ok? thanks :)

dannon:
I don't know how much it matters, but I think "Galaxy Server" repo is probably equal parts UI/UX and Backend

beatrizserrano:
also had doubts about Bioblend and LSP Server... maybe we could leave these 2 empty?

dannon:
If I had a question about bioblend or the API in general I'd ping the backend group, so that one makes sense.  LSP Server should technically probably be in the tools group's umbrella, but I know David is in the backend group and he's who I'd talk to about that.  So I don't have a good answer for that one :)

tnabtaf (approved):
@beatrizserrano @dannon 

1. Can we agree working group assignment is hard, and just do the best we can?  I say these are fine (and if they aren't we can change them)
2. Bea, can you change the WG link to conform to the guidline:
  * 🐣 Format links to other pages within the Galaxy hub with a leading `/src` and a trailing `index.md`.

beatrizserrano:
@tnabtaf fixed (I think)

dannon:
@tnabtaf Totally, I think this looks great -- I just wanted to have a pointer for folks looking to make UI/UX papercuts (which are most of the papercuts, I think, for the galaxy core repo).

PR #733, opened 2021-05-07
Add Github Action to optimize all images automatically

davelopez:
Hi!
We have just [tested this action](https://github.com/usegalaxy-eu/website/pull/662) on the EU website and seems to work quite well, see the result [here](https://github.com/usegalaxy-eu/website/pull/663).

I've made this PR in case you want to use it too :)

The first time it runs will probably create a PR with lots of updated images, and then it will only update when there are new images that can be optimized as part of a PR or periodically every Sunday at 11 PM.

Hope it helps!

### DIFF
--- a/.github/workflows/optimize-images.yml
+++ b/.github/workflows/optimize-images.yml
@@ -1,0 +1,56 @@
+# This Action will run in the following scenarios:
+# - on Pull Requests containing images (not including forks)
+# - on pushing of images to `master` (for forks)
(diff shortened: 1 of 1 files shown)


dannon:
@davelopez Great idea!  Thanks, let's see how it goes.

bgruening:
Uh does this work with git-lfs?

davelopez:
I hope so 😅 we will see when someone runs it manually or in the next PR with images on it.

dannon:
Good question, I didn't even think about LFS.  If it blows up, I guess we'll know :)

PR #754, opened 2021-09-02
Cancer informatics update blog t2 2021

luke-c-sargent:
title; let me know if using <div> in the markdown text to center images is disallowed magicks


tnabtaf (approved):
Thanks @luke-c-sargent. Looks good

PR #767, opened 2021-09-12
Add GitHub edit to gridhub branch

dannon:
Needs some more tinkering (see TODO) but it's a start.


dannon:
Merging this in; will follow up with more tweaks to how this is handled for aggregate pages.

PR #768, opened 2021-09-13
Add roscoff december training

abretaud:
Adding a new (in person!) training in december. It will be in french, but adding it anyway for the record. Feel free to adapt as needed


tnabtaf (approved):
Thanks @abretaud

PR #805, opened 2021-09-16
Omit empty jumbotron.

NickSto:
If `jumbotron.md` is present but empty, it would create a big, blank space on the homepage under the title.

This fixes that.

### DIFF
--- a/src/pages/Index.vue
+++ b/src/pages/Index.vue
@@ -4,7 +4,7 @@
             <h1 class="display-4">{{ $page.main.title }}</h1>
         </header>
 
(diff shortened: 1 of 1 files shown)


src/pages/Index.vue
```diff
@@ -214,6 +214,9 @@ query {
     padding-top: 100px;
     border: 4px solid black;
 }
+.header + .main-content {
+    margin-top: 40px;
```
dannon, line 218: @NickSto We should probably stick with rem here for margins?

NickSto: Done in #807

PR #806, opened 2021-09-16
Fix gitter sidecar.

dannon:
The problem was that sidecar would actually load before the mounted()
function that sets up the configuration was available, causing the
error.  This ensures a consistent order and less prone to error anyway.


dannon:
@martenson This should fix it up for you; thanks for reporting that.

martenson:
Thanks for fixing it!

PR #810, opened 2021-09-16
Improve the NavBar Github "edit me!" link

NickSto:
This should handle almost every page:  
All Collections  
\- Article
\- VueArticle
\- Platform  
Dynamic pages  
\- src/pages/*.vue

For dynamic pages, I decided to just point to the `main.md` file instead of the `.vue` file since I'd imagine people who are interested in messing with the structure already know where the repo is. Those who need the link are more likely to be interested in fixing errors in the text.

One exception I know about is the 404.vue page, which has no `main.md`. The url will be broken on that page.

Perhaps in the future pages could define their own `sourcePath` that NavBar then uses.


dannon:
@NickSto Superseded by #815, right?

NickSto:
@dannon Right, #815 and #811 have covered most of the territory addressed by this one.

The rest should go back into the ideas bin (now memorialized in #831).

PR #811, opened 2021-09-16
Simplify importing config.json - import it directly.

NickSto:
Didn't realize you could `require()` a json directly!

Much simpler.


dannon (approved):
Yep!

PR #813, opened 2021-09-16
Remove redundant `startswith()` util.

NickSto:
Apparently I wrote a `startswith()` function before I realized `String.prototype.startsWith()` exists!


dannon:
Heh, I didn't notice this ;)

PR #815, opened 2021-09-17
A more conservative improvement to the editUrl NavBar link.

NickSto:
A less radical change than #810.

This provides a specific Github url for every current Collection (`Article`, `VueArticle`, `Platform`), and leaves dynamic (`.vue`) pages alone.

It accommodates the different naming of the main `$page` object by basically duck typing: it looks for anything with a `.fileInfo.path` property.

It also doesn't assume anything lacking that is a dynamic page, so for those cases it provides a general link to the repo.

### DIFF
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -57,25 +57,38 @@
         </b-collapse>
     </b-navbar>
 </template>
+
 <script>
+import path from "path";
+import { rmPrefix } from "~/utils.js";
+import CONFIG from "~/../config.json";
 const REPO_URL = "https://github.com/galaxyproject/galaxy-hub/";
-const EDIT_URL = `${REPO_URL}tree/master/content/`;
+const EDIT_PATH = "tree/master/content";
 export default {
     computed: {
         editUrl() {
-            // TODO: a more robust way to do this, this only works on
-            // article-based pages and we probably don't want a hacky mess of
-            // exceptions.  For now, just default to root (which has a README,
-            // etc., so is not unreasonable) for 'special' pages.
-            let articlePath = this?.$page?.article?.fileInfo?.path;
-            if (articlePath) {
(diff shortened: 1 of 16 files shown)


src/components/NavBar.vue
```diff
@@ -57,25 +57,38 @@
...
-                articlePath = articlePath.replace(/^build\//, "");
-                articlePath = articlePath.replace(/^content-md\//, "");
-                return `${EDIT_URL}${articlePath}`;
+            // This will only point to the exact Github source url for Collections which include a "fileInfo { path }"
+            // in their GraphQL query. Otherwise, (like for dynamic pages) it'll default to pointing to the repo home
+            // page (which has a README, etc., so is not unreasonable).
+            let sourcePath = getPath(this.$page);
+            if (sourcePath) {
+                sourcePath = rmPrefix(sourcePath, CONFIG.build.dirs.md);
+                return path.join(REPO_URL, EDIT_PATH, sourcePath);
```
dannon, line 76: Probably a non-issue in reality, but we actually don't really want path.join using the platform separators -- we want explicit `/` joining these bits since it's a URL.

NickSto:
And actually, I realized this allows dynamic pages to opt-in by including a `fileInfo { path }` in their GraphQL query. And the page can even choose which Insert should be linked to. This actually makes it work for `404.vue`, which #810 didn't.

src/components/NavBar.vue
NickSto: touche, yeah

dannon: (i.e. -- if you were to run this on a windows machine you would actually get `\` slashes joining the path here)

NickSto: although it does nicely handle situations where there's already a "/" prefix or suffix on one of the parts, like in `REPO_URL`

dannon (approved):
Looks good!

PR #858, opened 2021-09-24
Add several new tests

NickSto:
NavBar, Bootstrap JS, redirects, static page rendering.

### DIFF
--- a/cypress/integration/main.spec.js
+++ b/cypress/integration/main.spec.js
@@ -6,9 +6,38 @@ describe("Main Page Tests", () => {
         cy.get("h2 > a[href='/news/']").should("be.visible");
         cy.get("h2 > a[href='/events/']").should("be.visible");
     });
+});
+
+describe("Sitewide tests", () => {
+    it("Visits the homepage", () => {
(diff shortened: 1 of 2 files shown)


dannon (approved):

cypress/integration/main.spec.js
```diff
@@ -22,37 +51,74 @@ describe("Test insert functionality", () => {
...
             cy.findByText(/Showing .+ total/i).should("be.visible");
         });
     });
 
-    /* 
     // TODO: This is the test that will fail until the /use/ page is fixed.
     // This will pass against a develop server currently.
-
-    it("Tries to visit another page after /use/, via the masthead", () => {
+    it.skip("Tries to visit another page after /use/, via the masthead", () => {
```
dannon, line 72: 👍 Better way to skip than big comment blocks!  I knew this in the back of my mind from jest land and didn't think to try it in cypress.

NickSto: Yeah, very satisfying to discover that and use it instead of a big comment.

PR #870, opened 2021-09-28
Bring run/build time Markdown parsing up to par with preprocessing parsing.

NickSto:
I've been using `mdToHtml()` in `utils.js` to parse bits of Markdown that aren't automatically parsed by the Gridsome transformers. For example: job summaries in `CardCareers.vue` or platform summaries in `Use.vue`.

But `mdToHtml()` was a thin wrapper around `remark-html`, an all-in-one Markdown-to-HTML parser. It lacked some of the features that `mdfixer.mjs` has: most importantly, `fix-links.mjs`, which removes the `/src` prefix and `index.md` postfix from links and makes image src paths relative.

In order to insert `fix-links.mjs` into the parsing pipeline, I had to replace the single-step `remark-html` pipeline with one broken down into 3 steps: Markdown parsing (`remark-parse`), Markdown to HTML transformation (`remark-rehype`), and HTML compilation (`rehype-stringify`). Then I could insert `fix-links.mjs` just after `remark-parse`.

Of course, these being [unified](https://unifiedjs.com/) packages, this wasn't that simple. Any reasonably recent version of `unified` and `rehype-stringify` is an ES import-only module, so I couldn't `require()` them into `utils.js`. Even [jiti](https://www.npmjs.com/package/jiti) didn't work. So I had to use the async `import()` function, turning `mdToHtml()` full async.

I was able to convert all its uses to async, though this means that the parsed Markdown only shows up in the browser when the page loads, not at build time.

Also, it causes a persistent issue on the `/use/` page where if you navigate away, then back, the parsed Markdown is replaced with plain text. It doesn't revert to the raw Markdown source; it's like the HTML tags get stripped out. At this point I think it might be an issue with the [b-table](https://bootstrap-vue.org/docs/components/table) component.

Oh, and `mdToHtml()` isn't able to make image links relative, since it lacks access to the source file's path. But I'm not sure this feature would be used anyway in any existing invocation. If it's important, we could probably add an optional argument to `mdToHtml()` and provide the path from the GraphQL object.


NickSto:
I'm closing this for now.

At the moment the only practical benefit this gives us is stripping `/src` and `index.md` from links in Markdown in metadata. I don't think there any instances of that left. And in any case, we've now [committed](863) to removing the `/src` and `index.md` permanently in the source, so we can just rely on that.

This still leaves us in a brittle state where all we get with the current `mdToHtml()` is the default parsing given by `remark-html`, and no easy way to modify it. So we may want to visit this PR in the future, if we find another need to augment the parsing.

PR #878, opened 2021-09-30
Added links to VGP video presented at Biodiversity Genomics

tnabtaf:


tnabtaf:
@NickSto Can you look at `homepage-videos.md` changes.  Is that an improvement?  If not, I will back it out.  If it is then I will apply to the other Video entries.

tnabtaf:
Whoa.  How does `homepage-pubs.md work`?  It does not have a `small` or a FontAwesome anywhere.  I suspect CSS...

NickSto:
Yes, the font size is set by [CSS](https://github.com/galaxyproject/galaxy-hub/blob/master/src/pages/Index.vue#L233), but the FontAwesome icons are hardcoded in the [template](https://github.com/galaxyproject/galaxy-hub/blob/master/src/pages/Index.vue#L73). Actually, the icon could be set in a field in the YAML metadata..

NickSto (approved):
Nice work!

PR #880, opened 2021-10-01
Modularize the homepage into HomeCard components, structure Insert data.

NickSto:
All the "cards" on the homepage have the same basic structure, which lends itself to a common component: the new `HomeCard`.

Structuring the Insert data into the YAML metadata instead of free-form Markdown means we can access it as a list of items, just like the recent events/news/blog/careers are a list of `Article`s. So `HomeCard` accepts a list of items, plus some metadata about how to make the header look.

You can still provide free-form Markdown if you'd like, which allows even the Twitter card to now be provided as an Insert.


dannon:
I merged up updated details here.  @tnabtaf look good to you?  And is this preferable to the free-form box, for your editing?

PR #930, opened 2021-10-14
add elixir CZ event

martenson:


beatrizserrano:
thanks @martenson, I'll mirror in EU :)

PR #939, opened 2021-10-15
/tutorials/var-hap/: Fix a couple links.

NickSto:
Fix a link to a local pdf and update a link to a Google groups post.


dannon:
@NickSto I merged this but it looks like it's got a broken link in the publish still -- that pdf isn't being handled right.  I can look tomorrow afternoon, but just FYI if you notice it.

PR #945, opened 2021-10-19
Update date COVID workshop

beatrizserrano:
@tnabtaf it would be nice to list New England BioLabs as well in the `location`, is that possible?


tnabtaf:
Yes, please add NEB.

Issue #979, opened 2021-10-27
Characters in hyperlinks getting escaped


NickSto:
## The Problem

Our current version of remark is breaking links which include characters like `&` and `_` by trying to escape them with a preceding backslash.

## Solutions

Currently, the only workaround is to use HTML for your hyperlinks instead of Markdown, which is super suboptimal.

The best solution would be to update remark, but it's Gridsome's dependency, so at the moment we're stuck with this version.

Another not-great workaround is to use Javascript to try to recognize these broken links and rewrite them.

## Examples
- [`/events/2021-09-papercuts/`](https://galaxyproject.org/events/2021-09-papercuts/) originally had a [link to timeanddate.com](https://www.timeanddate.com/worldclock/fixedtime.html?msg=14:00+Melbourne+Galaxy+Papercuts+CoFest+Call\&iso=20210916T14\&p1=152\&am=30) that got completely [munged](800).
- [`/news/2021-05-galaxy-update/`](https://galaxyproject.org/news/2021-05-galaxy-update/#publications) has (as of 54d4600e13) a doi link that's now broken: [https://doi.org/10.1007/978-1-0716-1307-8\\_20](https://doi.org/10.1007/978-1-0716-1307-8%5C_20)  
Actually it seems that's a little different: it only happens if it's a bare url in the Markdown, not an actual, explicit Markdown link. Once it's a Markdown link, the backslash disappears. But this doesn't seem to happen with ampersands.

Note: The original issue this came up in is #800, but I wanted to make this as a more general issue not just about that instance.

PR #1001, opened 2021-11-02
Prevent hyperlinks from getting broken by escaped characters.

NickSto:
This defines a custom compiler for Markdown hyperlinks. It ensures that when `mdfixer.mjs` fixes Markdown files during build preprocessing, `remark-stringify` doesn't escape characters in urls like `&` and `_`.

This fixes most cases of #979.


dannon:
Pushed another change reorganizing the tests a little bit into separate suites, which also fixes the imports in the middle of the .mjs file.

PR #1002, opened 2021-11-03
Run tests (link checking and jest) on PR submission.  Fix serveral (new) busted links.

dannon:
This updates our github workflows to run (internal only) link checking and jest unit tests on PR submission.  This caught several freshly broken links which are also fixed here.


dannon:
The tests failed, which means they actually worked, because they caught new broken links!

I'll fix the links here, too, because I might as well.

tnabtaf (approved):

PR #1010, opened 2021-11-08
Markdownlint dependency and first pass

dannon:
I need to spot check a bit of this, but it should eventually help get all our markdown in better shape.


NickSto:
I think some of these Markdown changes are pretty aggressive and could either change the meaning of the text or give us trouble in the future. I think we definitely need to sit with this a little longer before merging.

### Some examples

Removing `$ ` prefix from any line in code blocks:
[/admin/get-galaxy/](https://github.com/galaxyproject/galaxy-hub/pull/1010/files#diff-693763489e029a6f24f5dd21c4c76e985648391298cab1b6b6567777ae4a38c5L142)

Replacing tabs in code blocks with single spaces:
[/admin/reference-mafs/](https://github.com/galaxyproject/galaxy-hub/pull/1010/files#diff-7a1cbd8987694039373ed4fc8bcf6b32d4da32337b2cef1d41177c447f11fc28L79)
[/admin/data-integration/](https://github.com/galaxyproject/galaxy-hub/pull/1010/files#diff-a77f453ece2603aebf38702c804adb385b849fdf42ce4c6b15db90d66c6d6c69L60)
These in particular are `.loc` files, which are *supposed* to use tabs.

Changing numbered list indentation:
[/admin/internals/data-sources/](https://github.com/galaxyproject/galaxy-hub/pull/1010/files#diff-58ca57dab925f419225203e414eaa8c299874edf15092846f36d8a74ad7b9278)
This is an instance where the existing Markdown was certainly wrong, but this makes the wrong fix, and in the process obscures the original intent of the Markdown. These are sublists which should be *more* indented, not less. If we were to encounter this later it'd be hard to recognize that there even *was* an issue, let alone how to fix it.

NickSto:
I'd say maybe we could start with a smaller subset of rules enforced for now?

dannon:
Yep, this was just a starting point and is going to need thorough review and exceptions added -- zero rush at all on getting this in.  I'll put it in draft to reflect that.

NickSto:
This is excellent, thanks for breaking it out like this. I'll start looking through the non-trivial ones when I have time.

NickSto:
Okay I've finished looking through these. I pushed some fixes to these fixes and now I don't think there are any significant issues, with the exception of MD007.

To sum up our off-thread convo:
Auto-fixing MD007 creates too big a diff to sort through manually and there are cases where it correctly identifies an error but applies the wrong fix. I think that leaves us with a still-broken page but the original intent of the author is harder to see. You pointed out we still have the original intent in the git history for anyone who wants to go fixing. I'm -0 on it, so go ahead and merge.

dannon:
Resolved conflicts; merging.  Will take another cut some other time to get ahead of the new content included in covid site migration.

PR #1011, opened 2021-11-08
Add local link testing all-in-one command

dannon:
Slightly streamlines local link testing -- with this you can just do a `yarn run test:links` and test links locally.

Also changes check-links to *always* print the output report to the console -- not just on failure.


dannon:
@tnabtaf Merging this, but just fyi -- local linkchecking is `yarn run test:links`

PR #1012, opened 2021-11-08
Fix some existing Markdown issues on old pages.

NickSto:
Fixed some broken Markdown discovered when investigating #1010.


NickSto:
This should be pushed as changes to #1010 instead.

dannon:
Cherrypicked over, thanks @NickSto

PR #1032, opened 2021-11-17
/use/ Cypress test: Specify element to search for text.

NickSto:
I've been getting failures for the Cypress test "Tries to visit another page after /use/, via the masthead" test because `.findByText(/Galaxy Project Stats/i)` matches multiple elements: the `<title>` element in the `<head>` and the intended page title in an `<h1>`.

This tells it to look only in `<h1>` elements.

### DIFF
--- a/cypress/integration/main.spec.js
+++ b/cypress/integration/main.spec.js
@@ -75,7 +75,9 @@ describe("Use Page Tests", () => {
         cy.wait(1000);
         cy.findByRole("button", { name: /Community/i }).click();
         cy.findByRole("menuitem", { name: /Stats/i }).click();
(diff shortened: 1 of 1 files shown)


dannon:
@NickSto Interesting!  I haven't seen this one and I didn't expect findByText to grab the title, but +1 to the fix

NickSto:
...okay
Not sure what happened. Ran the all the tests myself and they worked. I'll look into it.

NickSto:
I cannot get it to fail on my end. I made sure to do a clean clone of the repo and everything.

Maybe this tweak will work, otherwise 🤷

Issue #1033, opened 2021-11-17
Headings can overlap floated elements


NickSto:
## Problem

Heading elements on content pages normally expand to fill the entire width of their container.

This can be an issue when there's a floating element which moves up and ends up alongside the heading. For example, before fd50bef, the headings on the left of [/events/gcc2013/program/](https://galaxyproject.org/events/gcc2013/program/) overlapped the linkbox on the right, making several of the links unclickable.

## Non-solutions

fd50bef tried to solve this by making headings inline-blocks, but then when there's several headings in a row, they bunch up next to each other on the same line.

#1024 tried to solve this in a different way by manually lowering the z-index of headings, but then some headings in special places (like popups) became invisible.

NickSto:
#1036 is a new attempt, which instead raises the z-index of the floated elements (all `.float-right`s). This should have fewer side effects.

This sets the z-index to 100. Popup elements' z-index is 1050, so this shouldn't wind up with linkboxes and images on top of popups.

NickSto:
Update: Looks like #1036 doesn't take care of all cases.

Example: Page [`/learn/faq/`](https://galaxyproject.org/learn/faq/) contains a linkbox which doesn't have the `float-right` class. Instead, it uses the specialized `linkbox` class, which floats it and does other styling.

dannon:
@NickSto Convert those two boxes to linkbox components?

NickSto:
Here are all the linkboxes I found which use the `linkbox` class:
```
$ git grep -E 'class.*linkbox[^-]'
admin/config/performance/linkbox.md:<div class='linkbox'>
admin/linkbox.md:<div class="linkbox trim-p">
admin/tools/linkbox.md:<div class='linkbox'>
datafiles/linkbox.md:<div class='linkbox'>
develop/linkbox.md:<div class='linkbox'>
documents/linkbox.md:<div class='linkbox'>
faqs/linkbox.md:<div class="linkbox trim-p">
learn/linkbox.md:<div class='linkbox'>
```
And who knows how many pages use these (since we'd have to edit every page that includes one of them).

We have higher priorities right now, so how about just adding a CSS fix for `.linkbox` elements for now?

We can keep an overall issue for converting linkbox Inserts to components later.

dannon:
Sure, #1038 should do it.

PR #1036, opened 2021-11-18
Bump z-index for float-right'ed things like linkbox

dannon:


NickSto:
This should fix #1033 in a way that hopefully doesn't break things.

PR #1038, opened 2021-11-18
specify z-index 100 on linkbox as well

dannon:


dannon:
xref #1033

NickSto:
I think this needs `position: relative` as well. See my updated comment: https://github.com/galaxyproject/galaxy-hub/issues/1033#issuecomment-973175362

NickSto:
@dannon Confirmed, position: relative is needed. I tried to push a fix but I'm still not able to push to your branch. Here's the patch:

```
--- a/src/assets/styles.scss
+++ b/src/assets/styles.scss
@@ -292,7 +292,9 @@ div.linkbox {
     background-color: #dfe5f9;
     font-size: 90%;
     text-align: center;
+    // Raise z-index to prevent linkboxes from being occluded by elements like headings (issue #1033).
     z-index: 100;
+    position: relative;
 }
```

dannon:
@NickSto That's weird.  I just checked, and this is selected:

image

And you're an admin, which is more powerful than a maintainer...  Maybe something on your end w/ branch configuration?

PR #1049, opened 2021-11-29
Add Cristóbal Gallardo to /iuc/.

NickSto:
Per galaxyproject/tools-iuc#4208

@gallardoalba welcome, and please double-check the info is correct


gallardoalba:
> Per [galaxyproject/tools-iuc#4208](https://github.com/galaxyproject/tools-iuc/issues/4208)
> 
> @gallardoalba welcome, and please double-check the info is correct

Yes, everything is ok. Thanks!

PR #1073, opened 2021-12-08
Convert COVID-19 logos footer into Markdown.

NickSto:
I got the logos to be a manageable size and the container element to be larger than its contents.

But I did it by making the logo heights uniform, while at their ideal sizes their heights are slightly different.
Also, there's not much spacing between the logos. I'm not sure the best way to go about increasing it.


NickSto:
Hey, turns out this fixes the #1075 issue in /covid19/! Probably best to merge it even if the display isn't perfect.

PR #1081, opened 2021-12-15
Usegalaxy welcome

dannon:
Has an unfortunate amount of html in the markdown here, but I converted what I could and kept the styling as close as possible.

Part of the difference here is that the current welcome uses Galaxy's exact stylesheet.  This is good motivation for addressing the style drift (however minor) between the two apps -- the core galaxy style knobs should all live in one spot.


natefoo:
FWIW I think if I link Galaxy's stylesheet in the welcome.html it applies to elements in the iframe.

natefoo:
Looks great to me locally.

PR #1093, opened 2021-12-17
GCC2022 update fixed(?)

NickSto:
Trying to fix #1091


NickSto:
Dannon did a better fix!

Issue #1075, opened 2021-12-09
Visiting certain urls directly results in broken navigation


NickSto:
This is the generalized version of #755. We squashed that instance, but this issue keeps popping up on other pages.

To reiterate the bug:

0. Build the site
    - This only happens in the fully built site, not the development server!
1. Visit the url of the problematic page directly.
    - Paste it into the url bar and hit enter or open it in a new tab.
2. Try clicking an internal link to another page on the Hub.
3. Watch the url change as intended and maybe some other things like the title or header image. But the main page content will not change.

Ideally we'd ferret out the root cause and prevent this once and for all.

But until then, this issue can keep track of pages where this is cropping up.

Note: This is also related to #758.

NickSto:
This is currently happening on https://galaxyproject.org/covid19/

NickSto:
#1376 is a very interesting example. It was caused by omitting the (normally optional) `<tbody>` tag in a table.

NickSto:
Noticed another set of pages where this is happening:
- [/events/gcc2019/](https://galaxyproject.org/events/gcc2019/)
- [/events/gcc2019/venue/](https://galaxyproject.org/events/gcc2019/venue/)
- [/events/gcc2019/training/](https://galaxyproject.org/events/gcc2019/training/)
- [/events/gcc2019/abstracts/](https://galaxyproject.org/events/gcc2019/abstracts/)
- ..etc

Those are the ones I manually checked, but I'm guessing it's happening on all GCC 2019 pages (and it's probably the header).

NickSto:
And another one: [/images/logos/](https://galaxyproject.org/images/logos/) (try clicking the "Learn" link).

PR #1096, opened 2021-12-19
Fix evil HTML introduced by former member of the team

tnabtaf:
I don't know about that guy...


NickSto:
Fixes the #1075 bug on /events/gcc2022/schedule/

PR #1102, opened 2021-12-23
The cleaning season has started!

beatrizserrano:
Done so far:
- Redirected tutorials to the GTN
- Simplified the learn page
- Restructured the masthead to make it more intuitive


beatrizserrano:
More changes: 
- Updated text in the home page
- Layout (only slightly), currently looking like this:
Screenshot from 2021-12-27 21-00-52

NickSto:
So I had a chance to check this out and it looks pretty good. I just have a couple thoughts:

1. Would it be possible to redirect each tutorial to a corresponding one at the GTN? I don't know how many have a direct equivalent there.
2. The NavBar looks good, though Anton might want to keep the "@jxtx" link.
3. I like the homepage redesign. It did feel like the jumbotron could take too much space sometimes. Though in the past we've had two main images side-by-side, so we might want to think about how the layout would work in that situation.
4. But I do think we could keep more of the homepage content in Markdown instead of the .vue file. I can work on that a bit.
5. The tests are currently failing because they depend on the current design of the NavBar. I can update them.

beatrizserrano:
Thanks a lot, Nick!

>     1. Would it be possible to redirect each tutorial to a corresponding one at the GTN? I don't know how many have a direct equivalent there.

That's my issue as well, I don't know which ones are equivalent in the GTN.

>     2. The NavBar looks good, though Anton might want to keep the "@jxtx" link.

I'll ask him.

>     3. I like the homepage redesign. It did feel like the jumbotron could take too much space sometimes. Though in the past we've had two main images side-by-side, so we might want to think about how the layout would work in that situation.

If we go for the option of changing the image every time the page builds, that wouldn't matter I guess.

>     4. But I do think we could keep more of the homepage content in Markdown instead of the .vue file. I can work on that a bit.
> 
>     5. The tests are currently failing because they depend on the current design of the NavBar. I can update them.

Thank you, I'll make a few more changes today and maybe we can merge, see how it goes and go for the next incremental step :)

NickSto (commented):
See comment in conversation thread

NickSto:
Hmm, not sure why the test is failing. It's working on my end.

NickSto:
D'oh, of course, it's cause those tests are for tutorials pages which now redirect to /learn. Should probably remove those.

NickSto:
Okay so for some reason this is now failing due to a number of broken internal links. I'm guessing many of these are pages you removed?

But it also reports the NavBar link to /tools/, so every page shows up in the broken link checker. Except when I try visiting /tools/ manually it works for me, so I'm not sure what's going on.

beatrizserrano:
Last version includes:
- Buttons per profile. @NickSto I've tried my best to create a component, but probably not clean code.
- Pages by profile (help wanted here with the content, I'm sure I'm missing plenty of resources).
- New images from https://undraw.co/.
- Links are fixed (I hope).
- Simplified navbar.

gif

NickSto:
I do have a couple style comments about the NavBar:

I feel like "@jxtx" being at the end, rather than "About", feels right, but that also could be because I'm used to it.

Also, this makes the NavBar significantly wider. This means there's a range of screen resolutions between 990 - 1200px wide where the NavBar starts wrapping and the display is messed up. I don't know how common these resolutions are, but it's not pretty. Here are some ways to fix it:
1. One solution would be to shorten the NavBar. Changing "Contribute!" back to "Edit" could do it, as could removing any one item.
2. We could also shorten the search box. Anything below 145px does it for me on Firefox. We could even use a media query to only shorten it for those screen resolutions.
3. Another possibility is changing the threshold at which it collapses into a hamburger menu to 1200px. I assume that's possible with Bootstrap. But the hamburger menu is far less usable so this is my least favorite.

NickSto:
Hmm, moving the search box below is an interesting idea. Maybe we put all the right-side stuff there. Ideally we could make it responsive so that it's all in one line when the viewport is wide enough, then two lines when it gets too small.

Oh, and one additional wrench to throw in the works: at a meeting this morning Anton mentioned that he'll want a covid link in the NavBar too once he's done writing those pages. I assume a top-level link.

beatrizserrano:
Thank you very much for all the help @NickSto!

PR #1108, opened 2022-01-07
Updated GCC2022 landing page to point to other GCC2022 pages

tnabtaf:


NickSto:
Note: This looks like it's causing (or surfacing) a11y test failures on the linkbox. We weren't seeing the errors before because there was no linkbox before.

The vertical pipes (`|`) that separate links are causing contrast errors because they're (intentionally) faint. An easy fix would be to darken them just enough.

PR #1126, opened 2022-01-17
Fix dates

nsoranzo:


beatrizserrano (approved):

beatrizserrano:
hmm the folder name should be also -03- instead of -01-. I have a local branch updating events, will fix there and upload changes. Thanks for noticing @nsoranzo!

PR #1129, opened 2022-01-18
Add redirect

hexylena:
@beatrizserrano is this right?


beatrizserrano (approved):
seems so! thanks for the fix @hexylena

PR #1165, opened 2022-01-27
Disbanding roundtables

assuntad23:
Created blog post
Removed future round table events
Removed section on Main Community Page
Added blurb to Devroundtable page with link to blog


assuntad23:
@beatrizserrano I removed all future roundtable events, including the Outreachy round table on the 17th. We need to re-create that event as a standalone special topic wg meeting.

assuntad23 (commented):
Thanks @dannon !

beatrizserrano (approved):
I like it @assuntad23, thanks!

dannon (approved):

afgane:
Having learned about this, I wanted to share my opinion in public too. Overall, in my opinion, canceling roundtables does not seem like the right move. It was a forum for presentation-style sharing of ideas and progress. Working groups are by definition meant to focus on work details and not presentations, making the entry point more difficult for non-WG members. Additionally, the WGs are much more siloed than a single, general event. 

Instead of just canceling, it would have been more transparent and inclusive to schedule a roundtable event and discuss its future. This comment is not intended at any one individual but a reaction to this organizational change to the project.

PR #1178, opened 2022-02-08
Fixes to welcome page

natefoo:
Turns out it's a good idea to actually view your changes before submitting them. Also I think the TACC logo should be centered but it's not?

xref #1177


dannon:
Whoops.  I can take an actual look this time :)

dannon:
@natefoo I fixed the tacc logo centering.   The text is imbalanced and we might want to reconsider how it's structured to even it out, but this is an improvement.

natefoo:
Thanks @dannon! I was just reusing style classes from who knows where.

PR #1236, opened 2022-03-16
create index.nd for next small scale meeting

hrhotz:


beatrizserrano (approved):
thanks @hrhotz for adding always your events :)

PR #1237, opened 2022-03-16
Change get started button to Get Started: First Steps with Galaxy

mariipia10:
Proposed change for Get Started button to make it clearer for newcomers where to go. Get Started could be understood as a instalation guide or "setting up your own galaxy instance" kind of guide.

image

@NickSto what do you think?


dannon:
@mariipia10 Thanks!  I pushed a commit with the results of `yarn format` run locally so the code quality test will pass now.

PR #1249, opened 2022-03-23
usegalaxy.org welcome: Restore the JXTX Foundation image now that Tapas is over

natefoo:


natefoo:
Close in favor of #1248

PR #1250, opened 2022-03-23
Add some basic infrastructure for subsites.

NickSto:
This adds the barebones start of the subsites feature for the Hub unification effort.

It adds a `subsites` metadata key, the `us` and `eu` subsites, events pages for both, and a method for customizing the styling of pages in different subsites.

With this merged, we'll already be able to mark events with different subsites and they'll appear on the proper events listings.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,7 @@
 {
     "host": "galaxyproject.org",
     "contentDir": "content",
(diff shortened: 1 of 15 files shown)


bgruening:
Wuhu! 🎉

PR #1251, opened 2022-03-23
Start categorizing posts by subsite

NickSto:
I thought I'd start tagging some events and news posts as `eu` or `us`, both to give the subsites logic some example data to work with and to start the discussion on what sorts of posts we want to categorize into specific subsites.

It was tricky to find clear examples of a "US" or "EU" post, since most events and news posts are trying for global reach. I don't particularly care if everything ends up being listed as global, so that's okay. From what I understand, the idea of making subsites in the first place was to allow us to have one, unified Hub while still letting people from different sites manage their own posts and content if they'd like. So we'll still maintain that option even if everything is global at first.

Also, a note: the posts I tagged as `eu` are basically duplicates also found on the current .eu Hub. When we eventually merge their content in, we'll remove these duplicates. And as a consequence, instead of the generic `eu` subsite, these events will maintain whatever list of EU-specific sites they currently have (`freiburg`, `pasteur`, etc).

### DIFF
--- a/content/events/2022-03-genome-assembly/index.md
+++ b/content/events/2022-03-genome-assembly/index.md
@@ -12,4 +12,5 @@ contact: "Bioinformatics Resource Center, Rockefeller University (brc@rockefelle
 tags: []
 links:
 image: ""
+subsites: [us, eu]
 ---
--- a/content/news/2021-08-elixir-update/index.md
+++ b/content/news/2021-08-elixir-update/index.md
@@ -3,4 +3,5 @@ title: "Updates from the ELIXIR Galaxy Community"
 tease: "An updated ELIXIR Galaxy website includes our goals, with a special emphasis on interactions with the local, domain-specific and global communities; plus much more"
 external_url: "https://galaxyproject.eu/posts/2021/08/24/new-elixir-galaxy-community-website/"
 date: "2021-08-24"
+subsites: [us]
 ---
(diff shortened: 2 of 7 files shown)


content/events/2022-03-genome-assembly/index.md
```diff
@@ -12,4 +12,5 @@ contact: "Bioinformatics Resource Center, Rockefeller University (brc@rockefelle
 tags: []
 links:
 image: ""
+subsites: [us, eu]
```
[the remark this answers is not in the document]
beatrizserrano: the location should be "US, Online", that's my bad, I probably copied it from another event

content/news/2021-08-elixir-update/index.md
```diff
@@ -3,4 +3,5 @@ title: "Updates from the ELIXIR Galaxy Community"
 tease: "An updated ELIXIR Galaxy website includes our goals, with a special emphasis on interactions with the local, domain-specific and global communities; plus much more"
 external_url: "https://galaxyproject.eu/posts/2021/08/24/new-elixir-galaxy-community-website/"
 date: "2021-08-24"
+subsites: [us]
```
beatrizserrano, line 6:
```suggestion
subsites: [eu]
```

content/news/2021-11-omicron/index.md
```diff
@@ -3,4 +3,5 @@ title: "Omicron and SARS-CoV-2 genome surveillance"
 tease: "A first view of the Omicron lineage’s mutational pattern derived transparently and fully reproducibly from raw sequencing reads"
 external_url: "https://galaxyproject.eu/posts/2021/11/29/omicron-and-galaxy/"
 date: "2021-11-29"
+subsites: [eu]
```
beatrizserrano, line 6:
```suggestion
subsites: [eu, us]
```

content/news/2021-08-elixir-update/index.md
NickSto: Oops, not sure why I had that as US.

Issue #1252, opened 2022-03-24
Broken link - Visualizations registry


elichad:
https://galaxyproject.org/visualizations-registry/ is broken.

Linked from https://github.com/galaxyproject/galaxy/blob/dev/config/plugins/visualizations/README.txt

These links under a search for "visualizations registry" also don't work.
![image](https://user-images.githubusercontent.com/20683271/159946723-5cd39a20-948e-495f-938b-295b017785fa.png)

beatrizserrano:
Thanks a lot for reporting @elichad! [This tutorial](https://training.galaxyproject.org/archive/2019-06-01/topics/dev/tutorials/visualization-generic/tutorial.html) seems to be a better resource for you to look at. Although it is still pointing to the same broken link (as "additional documentation"), the tutorial itself should be more helpful. However, if you still want to look at the old content, it's [here](https://github.com/galaxyproject/galaxy-hub/tree/e69ae2dcde30070518f8ecb35b976605ae200346/content/visualizations-registry).
I've created a [PR to update the link in the readme](https://github.com/galaxyproject/galaxy/pull/13616) and [an issue](https://github.com/galaxyproject/training-material/issues/3288) in the GTN to get the tutorial updated. Thanks again!

PR #1255, opened 2022-03-25
Add redirects to visualization-registry

beatrizserrano:
@NickSto have I done this right? Or left? ;) I am not sure if I needed to add the redirects to the subfolders too (code, configuration, cookbook).


NickSto:
Haha yeah that looks left!

Yeah, these metadata redirects are all on a per-page basis so it only takes effect for whatever page(s) you add the metadata to. So it's good you added it to all the subfolders.

NickSto:
CC #1252

PR #1263, opened 2022-03-30
Covid project: Remove client build instructions

wm75:
Guess that embedded markdown was just an accident.


dannon:
@wm75 Yeah, good catch -- that was the example used when we developed the markdown embed.

PR #1267, opened 2022-04-01
Fetch pubs from zotero on home instead of manual curation

dannon:
Before this is merged, we probably need to clean up publication details a bit on zotero and pick precisely which ones we'll use.

Right now I'm using:
```
                    {{ item.meta.creatorSummary }}, 
                    {{ item.data.journalAbbreviation ? item.data.journalAbbreviation : "preprint" }},
                    {{ item.data.date }}.
```

So, creatorSummary, journalAbbreviation, and date should be carefully added.  We can add more if we want.  It looks like journalAbbreviation isn't always correct or is sometimes missing currently, so we'll need to update those records.

Another option is to just hit crossref for each doi and request a proper citation format, as mentioned before.  This wouldn't be a lot of extra work, but I think we wanted to use this current approach to force zotero to be curated more.

Drops 'New Platforms' box as well, since we've decided to roll that into news when appropriate.

### DIFF
--- a/content/home/research.yml
+++ b/content/home/research.yml
@@ -1,19 +1,5 @@
 heading: Galactic pulse
 
-research1:
-    title: Recent publications
-    content: "Galaxy can be installed locally, fine-tuned with a tailored toolset and no usage quotas."
-    url: "https://www.zotero.org/groups/1732893/galaxy"
-    links:
-        - title: 'Dantan, L. et al. "Microbial education plays a crucial role in harnessing the beneficial properties of microbiota for infectious disease protection in Crassostrea gigas"  Sci Rep 14, 26914 (2024)'
-          url: "https://doi.org/10.1038/s41598-024-76096-4"
-        - title: 'Yujing Ren et al. "Effects of microplastics on litter decomposition in wetland soil" Environmental Pollution, Volume 343, (2024)'
-          url: "https://doi.org/10.1016/j.envpol.2023.123145"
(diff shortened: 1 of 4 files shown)


beatrizserrano:
Thank you, @dannon! I've updated the library to automatically get some authors that were missing (field `Creator`) so that should work fine now. However,  the `Journal Abbreviation` is not retrieved automatically (or I don't know how to do it) :(

dannon:
@beatrizserrano You know more than I do about Zotero :)  Is there another field that's more appropriate you'd like to use?

scottcain:
Oh, this would be so much better than https://github.com/galaxyproject/galaxy-hub/issues/2992

@bgruening along with me have been adding papers to Zotero on a fairly regular basis. We've been tagging them with the platform that the authors use (if it can be shown), so grabbing the first several where the tag is `>UseGalaxy.org` would be cool, as well as a page that has all of the papers in Zotero that have that tag (for galaxyproject.org of course)

dannon:
@scottcain How much do we care about the altmetric figures alongside the publication?  I know at one point we really wanted those, but currently this doesn't include them (though hasn't yet stripped out the scripts to make them).  We can obviously adjust styling and even change the type from 'donut' to something else, but they would be something like this:

image


My concern is that even if we get them looking great, the N most recent publications from Zotero (as we select them now) may not have altmetrics worth viewing.

hujambo-dunia:
I'm going to go ahead and merge this since the deadline is today - we can always improve later.

PR #1273, opened 2022-04-06
Removing Merged FAQs from the Hub

Melkeb:


Melkeb:
I have one concern regarding the FAQ titles. Should the FAQ title be the same as the GTN FAQ or is it okay if its different?

beatrizserrano:
> I have one concern regarding the FAQ titles. Should the FAQ title be the same as the GTN FAQ or is it okay if its different? (Most of them are renamed on the GTN FAQ)

I think that it's ok like you did it, and we can always edit if needed. Let's merge this so that you can move on with the next steps :) Thanks so much!

beatrizserrano (approved):

PR #1281, opened 2022-04-07
Updated Job states in Hub

garimavs:
I have changed the content of the FAQ in the Hub with the link in the GTN.


annefou (approved):
It looks good to me.

PR #1283, opened 2022-04-08
Add subsite UI to NavBar

NickSto:
This labels the current subsite in the NavBar, and allows theming from config.json instead of hardcoding it in styles.scss.

I tried to copy the current themes for the subsites on galaxyproject.eu.


NickSto:
This currently fails an accessibility test but I can't decipher the error message.

NickSto:
Also it looks like I'm getting links to `/null/news/` etc on the root subsite, but that's an easy fix.

PR #1286, opened 2022-04-10
Removing Moving Dataset between Galaxy Servers

Melkeb:
I was confused about removing this FAQ because it was a link to an external blog and addressed moving datasets and histories which I only addressed the datasets part but after exploring the GTN FAQs I found a similar FAQ about [moving histories](https://training.galaxyproject.org/training-material/faqs/galaxy/histories_transfer_entire_histories_from_one_galaxy_server_to_another.html) that addressed what was in the history section of the external [blog](https://genomicsvirtuallab.wordpress.com/2016/07/13/moving-data-between-galaxy-instances/). So, I renamed it "moving datasets..." and only linked it to the moving datasets GTN FAQ page. 

However, The link has also been mentioned multiple times throughout the FAQ page which I'm not sure if its necessary (I included all of it just incase) and another concern I had was on one of the mentioned parts it was linked to [another page](https://galaxyproject.org/blog/2016-07-moving-data-between-galaxy-instances/) which had the authors name and everything. I also replaced that page with the GTN FAQ link but I'm not sure if its better to leave the page as it is and only replace the link that is mentioned on the page that links to the external blog page.

### DIFF
--- a/content/support/download-data/index.md
+++ b/content/support/download-data/index.md
@@ -6,7 +6,7 @@
 Related topics
 
 * [Loading Data](/support/loading-data/)
-* [Moving data between Galaxy servers](/blog/2016-07-moving-data-between-galaxy-instances/)
+* [Moving datasets between Galaxy servers](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_moving_datasets_between_galaxy_servers.html)
 
 Tutorials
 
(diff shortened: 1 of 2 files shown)


beatrizserrano (approved):

beatrizserrano:
thanks @Melkeb!

PR #1290, opened 2022-04-13
remove small scale admins

hrhotz:


beatrizserrano:
I don't think that we should remove this but move it to the section `Grown-up Working Groups`. And the same for the other subcommittees that were discussed today at the meeting

hrhotz:
fine with me, see  #1292

PR #1297, opened 2022-04-15
Removing FAQ -  How to format fastq data for tools that require .fastqsanger format? 

Melkeb:


content/support/fastqsanger/index.md
```diff
@@ -13,23 +13,5 @@ Related Topics
 
 ## Help for Fastq Datasets
 
-### Fastqsanger format is usually required
+### [How to format fastq data for tools that require .fastqsanger format?](https://training.galaxyproject.org/training-material/faqs/galaxy/datatypes_fastq_format.html)
```
beatrizserrano, line 16: changing the text in this link would also break the link, I am not sure how to fix this but maybe we keep the old one?

```suggestion
### [Fastqsanger format is usually required](https://training.galaxyproject.org/training-material/faqs/galaxy/datatypes_fastq_format.html)
```

Melkeb: Ohh Okay. Yeah we can keep the title. Wouldn't deleting the other titles also break the link?

beatrizserrano (approved):

content/support/fastqsanger/index.md
beatrizserrano: yes, this will happen, but don't know how to solve it because the whole content is within the FAQ... any ideas?

PR #1298, opened 2022-04-15
Update GCC2022 logo on usegalaxy

afgane:
@beatrizserrano, we're down to ~170kb for the image.


beatrizserrano:
> @beatrizserrano, we're down to ~170kb for the image.

Wonderful! will update EU as well :) Thanks for updating and for the nice logo 🤩

beatrizserrano (approved):

PR #1300, opened 2022-04-15
NavBar: Add navigation between subsites.

NickSto:
This adds a dropdown menu that allows navigating between subsites, much like they have on galaxyproject.eu:

image

We probably don't want to merge this just yet for a couple reasons:
1. The links to the different subsites go to their homepages, which don't yet exist.
2. With this PR, the NavBar will always display the current subsite, even when you're on the "root"/"Global" one (i.e. all of our current pages).
   - First off, this means the entire subsites feature is prominently shown, when it's currently a mostly empty prototype.
   - Second, I'm not sure what people think of the "Global" label. Previously, the "root" subsite simply wasn't labeled in the NavBar. This feels more correct to me. But then where do we put the subsite dropdown? I'd love to hear thoughts.


NickSto:
Okay so the broken link checker is complaining, which makes sense - the NavBar contains links to the missing home pages of all the subsites.

bgruening:
@NickSto we could link to the currently hosted downstream pages until we move over the content.

For the naming a few ideas:

* sub-sites
* usegalaxy.*
* regions
* world-wide

But global is also good imho.

NickSto:
@bgruening Good point, we could just link to the current pages on galaxyproject.eu until they're migrated. Then there'd actually be a point to surfacing this now!

As for naming, the intention of the "Global" label in the screenshot is to tell you which subsite you're currently on. So if you're looking at an Erasmus page, this is how it looks:
image

So the question is whether we should even have a label there if we're on a "Global" page, and if not, where do we put the subsite dropdown? Or if we keep it, is there a better label than "Global"?

beatrizserrano:
Thanks @NickSto, this is cool :)

A couple of ideas:
- Since Europe is an aggregation of the current sub-sites in EU (I don't remember this well @NickSto), would it make sense to have a submenu with them?
- I think that having a "Global" dropdown would be useful, as for the naming:
  - Global: I like it, it's a nice statement, although it doesn't accurately tell what you'll find under that menu.
  - World-wide: similar to "Global".
  - Regions: this may be confused by the [regional communities](https://galaxyproject.org/community/#regional-communities).
  - usegalaxy.*: this may be misleading, thinking it's pointing to the servers?
  - Sub-sites: EU uses `Member Sites`, what about that one?

NickSto:
@beatrizserrano

> * Since Europe is an aggregation of the current sub-sites in EU (I don't remember this well @NickSto), would it make sense to have a submenu with them?

I was thinking a flat list would be more manageable than a multi-tiered set of dropdowns. I'm also not sure the `<b-navbar>` component can handle nesting. If people feel strongly though, I could look into what it'd take to do it.

>    * Global: I like it, it's a nice statement, although it doesn't accurately tell what you'll find under that menu.

I agree. An actual label for the dropdown (like .eu's "Member sites") would be better. At this point the issue is that we're running out of space to have both the label for the current subsite *and* a dropdown to select a different one. Even with the addition of this combined label/dropdown, the NavBar is basically full. We could start widening it, but I'm not sure at what point it's too unwieldy for anyone who doesn't have the window maximized.

beatrizserrano:
yes, space is definitely an issue... it's too crowded already. At some point I wanted to explore the option of having the NavBar out of the blue area, something like [this](https://colorlib.com/wp/template/publishing-company/). That would give us some extra space. Is it feasible? Also, for temporary items (e.g. GCC) we could have a nice button instead of the item in the NavBar

NickSto:
Hmm, you mean basically having a taller NavBar? Like, two rows? We could have the Galaxy logo, subsite label, dropdown, search bar, and edit button on one row and everything else on another. And since it's not a floating bar that always stays on-screen, taking up more room wouldn't be as annoying.

dannon:
I'd caution against shoving more stuff up there (and against making more room to do so by moving it to two layers or a level under).  It just gets to be too many links; instead, we probably need to just reorganize?

beatrizserrano:
Agreed (with both of you). The reason why I think that 2 levels (or 2 bars) would help is because we would have 2 subgroups of items: 
- logo + member sites + search bar + GH edit
- News + Events + ... 

But I don't have strong arguments to defend this, so let's do what you think it's best :) Either way, we need to clean up a bit, how do you suggest that we reorganize items in the current bar @dannon?

dannon:
@beatrizserrano It's a good question I wish I'd thought about more before suggesting!  At first glance, it does feel like the last 3 options (GCC2022, Covid19, and jxtx) are a different kind of thing from the rest.  I don't have great ideas to reorganize those, yet, though.

NickSto:
It feels like Training would fit naturally under Support, but maybe people want to keep it in a prominent, top-level place? The only other idea I have is putting News under About or Community. Arguably if you land on the homepage you're already shown recent News, so it's already advertised enough?

NickSto:
Okay, I've incorporated the feedback above:
- Training is now under Support.
- The subsite selector dropdown is now in a separate item named "Regions".
- The Global label is now displayed when on a Global page.

I've also adjusted the spacing to put a break between the "News" and "Events" links and the rest. This groups them with the subsite label, and puts the rest off to the right. This seemed appropriate because those links point to the subsite-specific news and events pages, not the global ones. I kept forgetting that those don't go to the global news/events pages, so this might help indicate that to users. It's not perfect, for sure, so I'm open to alternatives.

image

beatrizserrano:
I like it :)

NickSto:
The `build-and-test` check seems to be hung at the "Build static site" step. Not sure why. It worked fine in #1350; building only took 7 minutes there.

dannon:
Looks like constraining `max-old-space-size` will get the job done here.  Let me check through this and I'll merge it in.

dannon (approved):

NickSto:
For the record, it seemed that the build was running out of memory, then hanging. So the test sat "running" for the 6 hour time limit, then got killed. Testing it locally, it took 6821 MB of memory, which is close to Github's [7 GB limit](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). FYI, the current master only used 6082 MB. So this PR upped the memory usage significantly, so it makes sense that this is the one that started failing.

The `max-old-space-size` option constrains the amount of memory allowed in Node's heap, so lowering it makes Node spend more time garbage collecting, lowering peak memory usage.

PR #1306, opened 2022-04-18
Removing FAQ Format Help for Tabular/BED/Interval Datasets 

Beachyesh:
Does this look okay?


beatrizserrano:
> Does this look okay?

yes, it does, thanks a lot @Beachyesh!

beatrizserrano (approved):

PR #1310, opened 2022-04-18
Removing FAQ Custom Reference Genome

Nurzhamalyrys:
Hi! 
I am removing FAQ Custom Reference Genome


beatrizserrano (approved):
thanks @Nurzhamalyrys!

PR #1313, opened 2022-04-19
Removing FAQ Download Dataset collection using command line

Beachyesh:
I noticed that I had not removed the part about Dataset collections and it is included in the same FAQ

### DIFF
--- a/content/support/download-data/index.md
+++ b/content/support/download-data/index.md
@@ -33,21 +33,8 @@ Download datasets from Galaxy to your computer by clicking on the "disk icon" wi
 [Dataset downloading incompletely?](https://training.galaxyproject.org/training-material/faqs/galaxy/dataset_incomplete_download.html)
 
 
-## [Command-line option](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_download_using_command_line.html)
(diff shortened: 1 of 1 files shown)


beatrizserrano:
@Beachyesh I'm unsure about this change because the current link is: https://galaxyproject.org/support/download-data/#command-line-option 
If we change the title, it will become: 
https://galaxyproject.org/support/download-data/#command-line-option-and-dataset-collections

So if it has been used somewhere already, it will go to the top of the page, maybe not a terrible outcome, what do you think?

Beachyesh:
> @Beachyesh I'm unsure about this change because the current link is: https://galaxyproject.org/support/download-data/#command-line-option If we change the title, it will become: https://galaxyproject.org/support/download-data/#command-line-option-and-dataset-collections
> 
> So if it has been used somewhere already, it will go to the top of the page, maybe not a terrible outcome, what do you think?

Aah I see, I had made the decision to club together collection downloads with the command line download option because I thought it would make it easier to find everything about downloading datasets in one FAQ. It does go to the top of the page after the title change, but since we are redirecting them to GTN anyway, I don't think it is that bad of an outcome and I think we can make this change.

beatrizserrano (approved):

beatrizserrano:
change made then! Thank you @Beachyesh

PR #1314, opened 2022-04-20
Add events: comm calls + Anton's mini-symposium

beatrizserrano:
@NickSto included your suggestions and the dynamic page seems to work locally, thank you!


NickSto:
This is great! So glad to see the new system working.

PR #1318, opened 2022-04-22
Moved Faq Loading data to the GTN

bernandez:
Changes done to this pull request were  removing  the following Faqs under Loading data.
-  Get Data: Upload,  
-  Get Data: EBI-SRA, and 
-  Get Data: Upload tool used with FTP.


NickSto:
FYI, it looks like this deletes the entire page, not just the sections you mentioned.

PR #1327, opened 2022-04-25
Removing FAQ - Mismatched Chromosome identifiers and how to avoid them (1)

Melkeb:
Removing FAQs

- Finding BAM dataset identifiers
- Directly obtaining UCSC sourced *genome* identifiers

PR #1329, opened 2022-04-25
Removing FAQ - Mismatched Chromosome identifiers and how to avoid them (2)

Melkeb:
I'm not sure how to remove [this FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) on the Hub because

1. The FAQs in this page ([the Hub](https://galaxyproject.org/support/chrom-identifiers/)) has their own links on the GTN see PR [#1327](https://github.com/galaxyproject/galaxy-hub/pull/1327) and method 3 and 4 added on the [GTN FAQ]( https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) lead to the FAQ page on the Hub so I'm not sure if adding this GTN FAQ link will be the best choice. What do you think ?

2. Other than the methods I mentioned on point 1, [the FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) includes Intro section (Which has no link on the Hub) and the [Note Section](https://galaxyproject.org/support/chrom-identifiers/#a-note-on-built-in-reference-genomes) mentioned on the Hub. If adding the link is necessary, should I replace the note section with the GTN FAQ link (Which I did in this PR in order not to break any links if any) or should I add a title at the intro section with the GTN FAQ link and remove the note section (Since the FAQ includes intro section and mentions the methods. Feel like putting it at the end will not make sense) ?

**What do you think? Should this be removed?**

### DIFF
--- a/content/support/chrom-identifiers/index.md
+++ b/content/support/chrom-identifiers/index.md
@@ -5,9 +5,6 @@ title: Chromosome Identifiers in Reference Genomes (and other -omes)
 **[Back to Support Hub](/support/)**
 **[Troubleshooting Help](/support/#troubleshooting)**
 
-Methods described help to identify and correct errors or unexpected results linked to inputs having non-identical chromosome identifiers and/or different chromosome sequence content.
-
-**If using a Custom Reference genome**, the methods below also apply, but the first step is to make certain that the [Custom Genome is formatted correctly](/learn/custom-genomes/). Improper formating is the most common root cause of CG related errors.
 
 ## Find BAM dataset identifiers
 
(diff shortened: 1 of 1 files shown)


beatrizserrano:
> I'm not sure how to remove [this FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) on the Hub because
> 
> 1. The FAQs in this page ([the Hub](https://galaxyproject.org/support/chrom-identifiers/)) has their own links on the GTN see PR [#1327](https://github.com/galaxyproject/galaxy-hub/pull/1327) and method 3 and 4 added on the [GTN FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) lead to the FAQ page on the Hub so I'm not sure if adding this GTN FAQ link will be the best choice. What do you think ?

I'm not sure, either. Methods 3 and 4 are pointing to the Hub, so we can't replace that content. We could leave as it is until those methods are converted into GTN FAQs?

> 2. Other than the methods I mentioned on point 1, [the FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_chromosome_identifiers.html) includes Intro section (Which has no link on the Hub) and the [Note Section](https://galaxyproject.org/support/chrom-identifiers/#a-note-on-built-in-reference-genomes) mentioned on the Hub. If adding the link is necessary, should I replace the note section with the GTN FAQ link (Which I did in this PR in order not to break any links if any) or should I add a title at the intro section with the GTN FAQ link and remove the note section (Since the FAQ includes intro section and mentions the methods. Feel like putting it at the end will not make sense) ?

What you did looks correct to me. You can still add the title pointing to the same FAQ for the intro section if you like. I think it'd be good to keep the note section just in case it's linked from somewhere else, although it may look a bit weird/ugly (we can always restructure in the future).

Melkeb:
Okay. Added a title (header 1) with the link to the GTN FAQ.

PR #1337, opened 2022-04-27
Add an `all-eu` subsite shorthand.

NickSto:
We'll probably want to post things to all the EU subsites but not the US one relatively often. This saves typing out all the EU subsites.


bgruening (approved):
Nice

PR #1338, opened 2022-04-28
Add community call video

beatrizserrano:


beatrizserrano:
@guerler could you add the link to the slides here too?

PR #1360, opened 2022-05-05
Remove unused covid19 styles.

NickSto:
I couldn't find any evidence of any of the styles in `covid.styl` being actually used by the covid project. They're global definitions and some are quite generic (like `.footer`), so they can interfere with styling elsewhere. So deleting them would be nice.

FYI, my searches that turned up nothing:
```bash
$ git grep shieldlist content/projects/covid19
$ git grep feature content/projects/covid19
$ git grep paper content/projects/covid19
$ git grep footer content/projects/covid19
```
Since all the styles in the file depend on those classes, I concluded that nothing was using these rules. But please let me know if you notice something broken.


dannon (approved):
Yep, the new version of this dropped any uses of these old styles.

PR #1364, opened 2022-05-07
Implement usegalaxy.eu middle pane

NickSto:
This replicates the middle pane of usegalaxy.eu, currently hosted by the EU Hub at https://usegalaxy-eu.github.io/galaxy/

I decided on the path `/hosted/usegalaxy-eu/main/`, so we can host the other EU middle panes under `/hosted/usegalaxy-eu/` and the usegalaxy.org middle pane under `/hosted/usegalaxy-org/`.

The only thing missing is the carousel, which we're probably replacing anyway.

Also I just couldn't get the footer to be flush with the bottom of the page, so I omitted the `background-color` since that highlighted the gap.

### DIFF
--- a/content/eu/data-policy.md
+++ b/content/eu/data-policy.md
@@ -1,0 +1,7 @@
+---
+title: Our Data Policy
+---
+
+| Registered Users | Unregistered Users | FTP Data | GDPR Compliance |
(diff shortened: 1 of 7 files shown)


bgruening:
Cool! Thanks @NickSto

PR #1365, opened 2022-05-09
Better implementation of usegalaxy.eu middle pane via BareArticle and iframe.

NickSto:
#1364 implements the middle pane for usegalaxy.eu with a custom dynamic page, [`src/pages/hosted/usegalaxy-eu/Main.vue`](https://github.com/NickSto/galaxy-hub/blob/394dfd9c60784d8664ddb9c1bf3d3c23ba9f6b0f/src/pages/hosted/usegalaxy-eu/Main.vue). The dynamic page was necessary because the pane includes a list of recent news and events.

But multiple middle panes include this same list of recent news and events, and I'd have to make a custom dynamic page for each. Instead of repeating that code, I thought it would be useful to be able to make the dynamic lists its own independent element, embeddable via an iframe. This also lets us make all the middle panes static Markdown files. Then they can be composed using iframes and Inserts instead of all the custom Vue code and GraphQL queries to incorporate the news/events lists and special notices and footers.

In order to get a static page without the normal site UI (NavBar, footer), I made a new Collection, `BareArticle`, with its own template. These pages live under `/bare/`, so this changes the url of the middle pane from `/hosted/usegalaxy-eu/main/` to `/hosted/usegalaxy-eu/main/`.

One lingering issue is that the Gitter chat button is rendered twice, once inside the iframe, and once outside it. This issue is inevitable with the BareArticle collection, which we need anyway. I'm working on a fix.


bgruening:
@NickSto that all sounds cool, just keep in mind that this side is embedded into Galaxy as iframe (afaik)? Would that mean we have multiple nested iframes? Is that good or bad?

NickSto:
@bgruening I'm not aware of any issues with nested iframes aside from a little more memory usage - I think the same as if the page was loaded by itself? @dannon would surely know more.

I'm not a fan of iframes but this does simplify the implementation quite a bit. We could probably replace it with a Vue component, though in that case it'd be rendered once per page it's embedded in instead of once sitewide.

bgruening:
I have no clue, I trust you. Just wanted to raise this :) 

Thanks for all your work!

dannon:
image

PR #1367, opened 2022-05-09
Move Gitter chat embed from index.html to its own component.

NickSto:
This moves the Gitter chat embed out of the global index.html, so that we can avoid having it in certain pages (like BareArticles).

This addresses the double-gitter embed issue with #1365. Now, the BareArticle won't have it by default, so it won't appear in the inner BareArticle iframe (the news/events). But we *do* want it in the outer BareArticle, which can be done by including the `<Gitter />` component in its Markdown.

I think this avoids the race condition from #806, since it manually loads the Gitter JS only after setting the needed global variables. But it could certainly stand a second pair of eyes.

### DIFF
--- a/src/components/Gitter.vue
+++ b/src/components/Gitter.vue
@@ -1,4 +1,5 @@
 <template>
+    <div></div>
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


bgruening:
Bonus points if we can replace that with Matrix :-)

PR #1373, opened 2022-05-10
First draft for EU home page

beatrizserrano:
@NickSto here's my first draft, will make comments/questions inline with my doubts :)


content/eu/main.md
```diff
@@ -12,8 +16,48 @@ subtitle: The homepage of the European Galaxy community
 
 The **Galaxy Community** is actively involved in helping the ecosystem improve and sharing scientific discoveries.
 
+<br><br><br>
+
 <div class="row justify-content-center">
```
beatrizserrano, line 21: (None of the links of this section are correct, I'll need to create the other pages first)
Question: Can I create cards instead of buttons? I guess that means having Vue components within the markdown file...

content/eu/main.md
```diff
@@ -12,8 +16,48 @@ subtitle: The homepage of the European Galaxy community
...
+
+  <a href="./data-policy.md" class="btn w-25 btn-primary btn-lg" style="margin:1px">
+    Data retention police
+  </a>
+</div> 
+
+<br><br>
+
+
+<!-- Events, News, Tweets here -->
```
beatrizserrano, line 38: Events, News are now at the bottom of the page. Is there a way to move them up?

content/eu/main.md
```diff
@@ -12,8 +16,48 @@ subtitle: The homepage of the European Galaxy community
...
+
+| Supported communities with their own subdomain |
```
beatrizserrano, line 40: This would be ideally another button/card to make the home page simpler

content/eu/main.md
NickSto: FYI, I actually already added the data policy - you can see it at the bottom [this page](https://galaxyproject.org/bare/usegalaxy-eu/main/#our-data-policy). It's currently just an Insert, so you can't link directly to it, though. I can put on its own page, or insert it into the homepage if you'd like.

Also, I just added another insert slot below the news/events, so if you want stuff to appear below it, just put it in `extra.md`.

content/eu/main.md
NickSto: Do you mean a Bootstrap card like [this](https://getbootstrap.com/docs/4.0/components/card/#example)? All you need for that is just HTML with some special classes, like in the example code in that link.

beatrizserrano: Yes, like what we use in the general home called `HomeProfile`

NickSto: Cool, yeah then all you need is HTML. Take a look at the HTML in the Bootstrap example, or in [HomeProfile.vue](https://github.com/galaxyproject/galaxy-hub/blob/master/src/components/HomeProfile.vue) if you'd prefer that style.

beatrizserrano:
@NickSto I'd like to have the twitter feed at the bottom too, next to the news and events, but I don't know where to add it.

Besides that, I think I'm done with the first rough draft to start getting feedback from the EU team. So I think that we could merge this to have it online and facilitate the discussion, although I wouldn't link it from the Home page yet. What do you think?

NickSto:
FYI: I've been using the `content` label for PRs that only touch Markdown (so I can filter it out)

beatrizserrano:
> Cool, that sounds great! Do you want to add the Twitter feed in a card next to the News and Events ones?

Yes, I think that the same format as in the global home page would work. Otherwise there is a weird empty space there.

> Oh and FYI: I've been using the `content` label for PRs that only touch Markdown (so I can filter it out)

Oh ok, sorry, didn't realize. Good to know :)

NickSto:
I'm going to add the Twitter card in a separate PR since that's a little more complicated.

PR #1402, opened 2022-05-18
Force more modern sharp for m1 mac compatibility

dannon:
Added benefit, the image processing step should be about twice as fast.  42 seconds to 19 on my machine.


dannon:
@tnabtaf If you `rm -rf node_modules` and install/build on this branch it should work cleanly.  Tested on @astrovsky01's new m1 here.

PR #1425, opened 2022-06-03
Add a Projects menu to the NavBar

wm75:
For now add the Covid19 and the MPXV efforts to it.
The idea is that projects listed under https://galaxyproject.org/community/wg/#current-set-of-projects should all have hub pages, then be added to the menu.


wm75:
@beatrizserrano do you think this makes sense? At least more than the current state?

beatrizserrano:
Yes, it does :) We discussed in the past to have also VGP under that menu whenever a page is available for the project

beatrizserrano (approved):

Issue #1428, opened 2022-06-03
More inclusive Projects navbar dropdown


wm75:
In https://github.com/galaxyproject/galaxy-hub/pull/1425 I added a "Projects" dropdown menu to the hub's navbar. My original idea was to give the projects listed under https://galaxyproject.org/community/wg/#current-set-of-projects (and for now our new monkeypox efforts) some visibility.

Now just saying "Projects" there and not listing community projects there, doesn't feel very inclusive (as pointed out by @bebatut :heart:).
What would be better (I think) is a dropdown with sections:

*Core Teams Projects*
 - Cancer Informatics
 - Human Genetics
 - COVID-19
 - VGP

*Community Projects*
 - Climate Science
 - ...

and have the section titles link to overview pages (like the one suggested by @bgruening in #1426).
The Core Teams Projects overview page could then also explains what exactly defines such a project.
What do you all think?

Now adding those section titles and turn them into links is beyond my very humble vue skills, but maybe someone more qualified can jump in?

dannon:
+1, of course, we just need to have pages for those projects.  Is there a more comprehensive list of what we can currently include anywhere?

I'm happy to mock up the links for ya if that'd help.

wm75:
@bebatut what would you include under *Community Projects*?

PR #1429, opened 2022-06-03
Add initial expanded projects list with links to overview pages

dannon:
xref #1428

### DIFF
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -46,8 +46,23 @@
                         <b-dropdown-item to="/images/galaxy-logos/">Branding</b-dropdown-item>
                     </b-nav-item-dropdown>
                     <b-nav-item-dropdown text="Projects">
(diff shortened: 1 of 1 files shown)


wm75:
Thanks @dannon - that's exactly what I had in mind!
@beatrizserrano @bebatut I will try to work on the "Core Team Projects" landing page above. Maybe you could start thinking about what a corresponding "Community Projects" landing page could look like? Maybe we really want to put the current not-so-visible "Communities of Practice" section there? In principle, we could list any community in the new dropdown menu that has somebody willing to create a hub page for them?

beatrizserrano:
I think that the "Communities of Practice" deserve a more prominent place, but I don't see them as projects. My understanding is that those communities meet regularly because they share an interest, not because they work on the same project. We could have a separate page for the communities of practice and another for the communities by location and add them under the "Community" menu instead of "Projects". And we could expand that information for those communities as much as the communities like, e.g. adding the different projects in which the members are involved.

PR #1468, opened 2022-06-22
Add some local temp files to .gitignore.

NickSto:
These are the names of some debugging files I often have lying around. They seem generic enough they might be useful for others?


NickSto:
Hearing no objections, I hereby merge!

PR #1504, opened 2022-06-30
Add gitpod configuration

dannon:
Enable gitpod configuration.

Allows editing the hub and creating PRs without any local install at all:

image


dannon:
Going to just go ahead and add this in the interest of making it available and we can add a tutorial later.

PR #1513, opened 2022-07-06
Remove prioritization section

guerler:
Currently our CoC openly discriminates and disadvantages people by race, gender and sexual orientation with regard to protecting their rights, this PR resolves this. Please also see: https://www.findlaw.com/employment/employment-discrimination/reverse-discrimination.html and https://www.eeoc.gov/youth/racecolor-discrimination-faqs.

### DIFF
--- a/content/community/coc/index.md
+++ b/content/community/coc/index.md
@@ -45,13 +45,6 @@ The Galaxy Community strives to:
   - Continued one-on-one communication after requests to cease.
   - Deliberate “outing” of any aspect of a person’s identity without their consent, except as necessary to protect vulnerable people from intentional abuse.
   - Malicious publication of otherwise private communication, or private personal information (“doxing” or “outing”).
-  - ‘Reverse’ -isms, including ‘reverse racism,’ ‘reverse sexism,’ and ‘cisphobia’.
-
-The [Galaxy CoC Committee](#the-galaxy-coc-committee) reserves the right not to act on complaints regarding:
-- Reasonable communication of boundaries, such as “leave me alone,” “go away,” or “I’m not discussing this with you.”
-- Communicating in a ‘tone’ you don’t find congenial.
-- Criticizing racist, sexist, or otherwise oppressive behavior or assumptions.
-
 
 ### Reporting
 
--- a/content/community/coc/index.md
+++ b/content/community/coc/index.md
@@ -45,13 +45,12 @@ The Galaxy Community strives to:
   - Continued one-on-one communication after requests to cease.
   - Deliberate “outing” of any aspect of a person’s identity without their consent, except as necessary to protect vulnerable people from intentional abuse.
   - Malicious publication of otherwise private communication, or private personal information (“doxing” or “outing”).
+  - ‘Reverse’ -isms, including ‘reverse racism,’ ‘reverse sexism,’ and ‘cisphobia’.
 
-
-The Galaxy Community prioritizes marginalized people’s safety over privileged people’s comfort. The [Galaxy CoC Committee](#the-galaxy-coc-committee) reserves the right not to act on complaints regarding:
-- ‘Reverse’ -isms, including ‘reverse racism,’ ‘reverse sexism,’ and ‘cisphobia’.
+The [Galaxy CoC Committee](#the-galaxy-coc-committee) reserves the right not to act on complaints regarding:
 - Reasonable communication of boundaries, such as “leave me alone,” “go away,” or “I’m not discussing this with you.”
 - Communicating in a ‘tone’ you don’t find congenial.
-- Criticizing racist, sexist, cissexist, or otherwise oppressive behavior or assumptions.
+- Criticizing racist, sexist, or otherwise oppressive behavior or assumptions.
 
 
 ### Reporting


tnabtaf:
Hi @guerler, all,

Sam, thanks for posting this suggestion.  I actually have a history with this clause (lines 50-54), and have given it some thought before.  But your post caused me to think about this more deeply than I have before, and that has shifted my opinion.

My very long-winded response to Sam's very pithy PR is structured as:

1. My history with this clause
2. Cultural versus Legal
3. Discussion
4. What to do?

People can skip the first section (my history) - it's just my attempt to establish street cred on this issue.  The rest is still verbose, but is more relevant.

## 1. My history with this clause

One of my initial tasks at my new job was to update the [conda community's code of conduct](https://github.com/conda-incubator/governance/blob/master/CODE_OF_CONDUCT.md) and get it formally approved.  The conda CoC is based on the NumFOCUS COC template.  That is now several years old, and that's a long time in CoC world.  Before proposing any updates to the conda COC, I looked at other communities' CoCs, including the recently remodeled Galaxy CoC.  

I remember coming across this clause in the Galaxy CoC and thinking something like:

> That's interesting.  I know exactly what situation this clause is addressing (see Cultural versus Legal below). Should I propose this for conda?

And I decided not to.  But not by very much.  I was open to it, but I didn't think it was necessary, and despite my verbosity here, I lean hard towards shorter governance documents.

But then another member of the conda CoC committee independently proposed that we add a clause like this one.  They hadn't looked at the Galaxy CoC. No one, including me, had strong objections so we added it.

And now your post is causing me to think about this in a much deeper way.

## 2. Cultural versus legal.

The links in the PR description cover the legal meaning of reverse discrimination in the US.

But, [in the US at least](https://fivethirtyeight.com/features/how-white-victimhood-fuels-republican-politics/), the legal definition is different from what most people think when they hear the words "reverse discrimination."  Rather it is most often used when people find the status quo being threatened, and they are attached to the status quo.

I believe that Codes of Conduct in general, and Galaxy in particular, are not about preserving the status quo, but are rather about protecting our community.
  
I believe the point of this section of the CoC is:

1. To discourage people from submitting this type of spurious case. 
2. To provide the committee some cover when people do submit this type of case.

## 3. Discussion

Note that "protecting the community" is not in that list.  I think this section is clearly about enabling the COC committee to avoid dealing with spurious cases.  Being on a COC committee is a really challenging task, but the impact the committee has on the quality of the community is enormous.  Spurious cases like these add to the burden of the committee *without resulting in any benefit to the community*.

The down side of this section is that it will discourage some people from reporting actual harassment.  For example, I work in community management, and I am often the only male in the room.  It is plausible that my contributions would be dismissed by others because I am male and "men don't understand people as well as women."  Note that I can't remember ever being dismissed/ignored because of my gender. (How many women can say that?)

If that happened to me, I would find the Code of Conduct, and then see this section.  And, 60 year old me would be irritated by it (having just experienced the situation being described), but I would still report the incident.  I would be worried that I wouldn't be heard, and that's a problem.
  
But what about a 30 or 25 year old Dave?  That Dave would be far less likely to report the incident, and that's a problem too.

Of course there is another option here: Maybe I felt ignored because I subconsciously expect that a roomful of women should inherently pay special attention to my opinions.  Maybe several ideas were presented, and mine was not embraced because it actually wasn't the best idea.  Maybe, if a roomful of men had picked another idea, I would feel irritated, but not persecuted because of who I am.

The challenge is that individuals may be poor judges of what is actually happening.  We are just too close to it.

## 4. What to do?

We could leave the current text in, and possibly discourage some valid reports.  Or, take the text out, and increase the number of spurious reports. 

I find that when I put it that way, I want to take the text out. I would rather err on the side of getting too many reports instead of too few.  This would increase the unpleasantness for the committee, but the committee's job is to protect the community, even when it is difficult.

I am not in favor of merging the PR as it now stands.  Explicitly listing “reverse-isms” under the partial list of example harassment would lead to exactly what this clause is seeking to avoid.  I think we should either keep lines 50-54 as is, or drop the whole thing.  Also dropping “cissexist” from the last line sends exactly the wrong message to our community.  We can always discuss how long partial lists like this should be, but once something is in a list, it should stay there. Dropping just one item seems to say “we don’t actually care about this community.” 

But (there is always a but), I am not on the Galaxy CoC committee, and you all should be leery of opinions from non-members that will increase the committee's work.  I'm also not absolutely certain in my opinion here, but as of right now, I expect to propose that we drop the similar clause from the conda CoC when we next review it.

Thanks for reading this far.

tnabtaf:
@guerler and I are both at the GCC2022 CoFest and we just discussed my comments.  And my comments are waaaay long.

So, summarizing my suggestion for the CoC here:

* Consider dropping everything in lines 50-54 because I believe the overall effect is to discourage people from reporting.  In general, I think the COC should encourage rather than discourage reporting.

guerler:
Thanks @tnabtaf for looking into this. I think thats a good idea too. Can you please check if I understood your suggestion correctly and its reflected in this PR now. Thanks again.

tnabtaf:
Yep, that is what I am suggesting. Thanks Sam.

beatrizserrano:
Thanks @guerler and @tnabtaf for the time and effort that you put on this.

I just wanted to give some updates about the ongoing process. Part of the committee has discussed your proposed changes, but some members are on vacation right now or are going to be very soon. So I can't speak on behalf of the committee in this message, I can only provide my personal view.

We still need to set up the procedure for updates and, as we told you, we will open such procedure for the whole community to contribute with ideas and improvements. Since this procedure has to be thoughtful and treated carefully, I think that it is important that most of the members of the committee are involved in its development. Therefore, I don't think that we can make a decision before September. So I apologize for that.

More specifically, here's what I propose:

1. The committee meets in September.
2. We set up a procedure for updates.
3. You (and anyone from the community) will be more than welcome to contribute to the discussion, as we did when we updated to the current version. We will leave it open long enough to give everybody the chance to participate.
4. You (and anyone from the community) will be invited to become part of the committee.
5. The committee writes a report to update the community about:
- the agreed changes,
- the version of the old and new CoC,
- an aggregation of the cases reported up to that point.

I appreciate your patience, thanks again.

jdavcs:
I don't know enough to discuss neither cultural nor legal considerations that may be relevant to this issue. My opinion is purely subjective. I read the Code of Conduct for the first time as a result of this discussion. The way I interpreted that section (and that's how I would've understood it had I read it any other time) is this: it is not about encouraging/discouraging spurious reports. It is about this: "The Galaxy Community prioritizes marginalized people’s safety over privileged people’s comfort."  The rest is clarification of how we act on it. I think, if we agree with this (as a community? as the Galaxy team?), we may want to explicitly state this in our Code of Conduct - as it is stated now.

annefou:
I fully agree with your comment @jdavcs and I am just adding that to me this paragraph is a step towards the greater equity and inclusion that the Galaxy Community is thriving at achieving. To learn more about it see Melke's blogs and in particular the one where she defines these [key concepts](https://galaxyproject.org/blog/2022-07-01-galaxy-wellbeing-blog-2/).

Equity versus equality

guerler:
I think, that if we insist on leaving the term `privileged` in the text, it needs to be further specified such that they can be clearly identified.

NickSto:
I think @jdavcs has pinpointed the key sentence in this section:

> The Galaxy Community prioritizes marginalized people’s safety over privileged people’s comfort.

One goal of the CoC is to provide mechanisms to deal with incidents, but another goal is to communicate to people that this is a safe and welcoming community. So I'd prefer to keep something like this in, rather than handle it post-reporting, as @tnabtaf suggested.

And the diagram @annefou shared is a great visualization of the goal of this section. Perhaps this section could make that distinction: the goal is not to disadvantage any group, but to acknowledge existing disadvantages and consider incidents in that light.

guerler:
Closing this, since the vote is not in favor. However closing this, or not, does not affect the issue at hand.

PR #1524, opened 2022-07-11
fgbs blog1

nekrut:


content/blog/2022-07-FGBS1/index.md
```diff
@@ -0,0 +1,20 @@
...
+tease: "Dealing with misconceptions one blog post at a time"
+authors: "The Galaxy Team"
+date: "2022-07-11"
+---
+
+![galaxy hops, super smooth](hops.jpg)
+
+Galaxy is a free open platform designed to give researchers across the world unconstrained access to tools and infrastructure necessary for analysis of their data. Galaxy does a lot of things and contains many components, making it difficult to quickly comprehend everything that it can do. It also evolves [rapidly](https://docs.galaxyproject.org/en/master/releases/index.html). As a result, there are a number of misconceptions about what it is and what can (or cannot) be done with it. To spread the ultimate knowledge across the Universe, we are starting the **“Fighting Galactic Bullshit (FGBs)”** series (yes, we were definitely inspired by a certain [effort](https://www.callingbullshit.org/)). This series identifies the most common misconceptions and explains why they are, simply put, BULLSHIT! It will include the following “episodes”:
+
+1. Galaxy is only useful for genome scientists…, **But, all scientists have genomes, right?**
```
mvdbeek, line 12: I guess that's the point of the blog post to come, but seems very cryptic. Not sure we'll be shooting ourselves in the foot with genome and dbkey plastered all over the interface.

content/blog/2022-07-FGBS1/index.md
```diff
@@ -0,0 +1,20 @@
...
+2. Galaxy does not scale to large and complex problems…, **Define large and complex!**
+3. Galaxy offers nothing to informaticians who can write their own code…, **Are you sure?**
+4. Galaxy is hard to learn and hard to use…, **And bicycles are hard to park!**
+5. Galaxy is only useful for teaching, not real science…, **Who are bigger snobs: teachers or scientists?**
+6. Galaxy is neither popular with, nor widely used by real scientists…, **It depends on the meaning of the word “is”.**
```
mvdbeek, line 17:
```suggestion
6. Galaxy is neither popular with, nor widely used by real scientists…, **It depends on the meaning of the words “popular” and "real".**
```
Again, not sure what will be in the blog post, but I feel like the meaning of `is` is not up for debate ?

bgruening (approved):

content/blog/2022-07-FGBS1/index.md
bgruening: It's about __only useful__ for genome scientists. The first blog post will cover this topic and will point to other communities using Galaxy.

mvdbeek: Sure, but what are we alluding to with `But, all scientists have genomes, right` ?

content/blog/2022-07-FGBS1/index.md
hrhotz: I like this 'clarification'

content/blog/2022-07-FGBS1/index.md
bgruening: I see your confusion. But this refers to "the plant next to me has a genome" :)
If that is too confusing we might need to find a new word here.

hrhotz: suggestion: 
What is a genome scientist anyway?

content/blog/2022-07-FGBS1/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: "Fighting Galactic Bullshit 1"
```
mvdbeek, line 2: I think we should rebrand this a little. I'm not so sure about referring to the misconceptions people have as " bullshit". It feels a little aggressive and some people may perceive this as being blamed. I'm worried that we're annoying people for no reason, and it also seems a little desperate and makes it sound like we're struggling, which is far from the truth. 

I think it's better to create a positive context when promoting something instead of focusing on negative things. So instead of saying "we're here to correct the BS" maybe we can instead highlight the progress we've made since the start 15 years ago, which is gigantic! We can point out several features that have been added since some people first encountered galaxy, which they may not have had the chance to hear about. That should already go into the introductory text. This would give a solid framework for most of the blog posts to come.

content/blog/2022-07-FGBS1/index.md
nekrut: you guys do not remember cultural references to bill Clinton trial, do you? 😄

content/blog/2022-07-FGBS1/index.md
nekrut: Most misconceptions are due to the fact that people simply do not know (similarly to how we sometimes do not know and make claims like "only synthetic oil is good for your engine"). But this is what bullshit is - normal oil is just fine if you do not have a turbo (or crazy compression).

content/blog/2022-07-FGBS1/index.md
hrhotz: ah! now it rings a bell ;)

content/blog/2022-07-FGBS1/index.md
mvdbeek: I'm saying this is cryptic. Is the joke that because everyone has a genome everyone is a genome scientist (your comp-sci employees, all philosophers, and Object Orientation would like a word with you 🤣 ) ? Or is it that it's irrelevant that Galaxy started that way ?

content/blog/2022-07-FGBS1/index.md
mvdbeek: Bullshit implies (to some degree) that if you believe this, you've not been smart enough to understand. If you don't want to change the direction I'd be in favor of replacing Bullshit with the more boring `Misconceptions`.

bwlang: I think Bullshit refers to the arguments of people who don't care about true vs. false - just what is in their interest (see https://en.wikipedia.org/wiki/On_Bullshit).  I'm not sure that applies here.  I suspect that most people who hold/repeat some of the perceptions being discussed actually think they are true and care about what is true (so hopefully could be convinced otherwise).  Misconceptions seems like a serious and academic word to me - maybe best used later in the post. I wonder about something inspired by clickbait practicioners like "N surprising things that Galaxy is good for".   On the other hand - bullshit certainly catches the eye... we need some more marketing expertise!

bgruening:
I guess we can close this one?

PR #1539, opened 2022-07-16
GCC2022 rejig layout for the event

afgane:


afgane:
@NickSto can you please explain what should be done about the failing test? https://github.com/galaxyproject/galaxy-hub/runs/7367243282?check_suite_focus=true#step:9:92

(it'd be nice to merge this asap to have the site ready as GCC gets underway so if it's not a big deal to fix later...)

NickSto:
It looks like both issues are with the "See full schedule" button.

One is a "color-contrast" violation:
> Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds
> Elements must have sufficient color contrast
> https://dequeuniversity.com/rules/axe/4.3/color-contrast?application=axeAPI

I think the orange is too close to the blue of the text. Maybe lightening it would fix it.

The other is a "nested-interactive" violation:
> Nested interactive controls are not announced by screen readers
> Ensure interactive controls are not nested
> https://dequeuniversity.com/rules/axe/4.3/nested-interactive?application=axeAPI

I think it's talking about the fact that the `<button>` contains an `<a>` link. You could just merge the two and let the classes (`btn btn-secondary center`) style the link as a button.

NickSto:
Btw you can test this interactively yourself before pushing by running `yarn develop` in one window, then `yarn cypress:open` in another. Then click `accessibility.spec.js` to run the a11y tests. That could speed up the troubleshooting process.

afgane:
Thanks for the explanation and the pointers @NickSto.

PR #1542, opened 2022-07-17
FTP decomission notice: emphasize that it's only the FTP service.

NickSto:
When I went to Galaxy, for a second my brain saw the big red notice and it looked like it was saying Galaxy itself was being shut down.

This bolds the word "FTP" as a feeble attempt to reduce the likelihood of that happening.

### DIFF
--- a/content/usegalaxy/welcome/main.md
+++ b/content/usegalaxy/welcome/main.md
@@ -4,7 +4,7 @@ title: usegalaxy.org
 
 <div class="alert alert-danger trim-p mt-2">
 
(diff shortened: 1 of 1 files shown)


mvdbeek:
Also should probably do this in a non-error color (blue / info) ?

NickSto:
@mvdbeek Hey, that sounds like a good idea

NickSto:
I switched it to blue. I'm not sure if that's too quiet now? The text is still pretty big and loud so it seems okay.

PR #1551, opened 2022-07-21
Community Directory WIP

NickSto:
@assuntad23

### DIFF
--- a/content/people/assunta/index.md
+++ b/content/people/assunta/index.md
@@ -1,0 +1,6 @@
+---
+name: Assunta DeSanto
+subsites: [us]
+github: assuntad23
+affiliation: psu
+---
--- a/config.json
+++ b/config.json
@@ -3,7 +3,8 @@
     "contentDir": "content",
     "collections": {
         "BareArticle": { "path": "/bare/", "type": "vue" },
-        "Platform": { "path": "/use/", "type": "md" }
+        "Platform": { "path": "/use/", "type": "md" },
+        "Person": { "path": "/people/", "type": "md" }
     },
     "categories": {
         "/blog": "blog",
--- a/content/people/0example/index.md
+++ b/content/people/0example/index.md
@@ -1,0 +1,20 @@
+---
+name: Person
+subsites: ["eu"]
+email: thing
+phone: asdf
+website: sadf
+github: asdf
+orcid: asdf
+gitter: asdf
+matrix: asdf
+linkedin: asdf
+twitter: asdf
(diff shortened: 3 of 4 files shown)


NickSto:
Note: There might currently be an issue with Markdown images in Person.vue.
Example: Look at /people/delphine-lariviere/ in this branch vs. [the live version](https://galaxyproject.org//people/delphine-lariviere/).

annefou:
Awesome to have the possibility to create individual pages! Great initiative! We (as part of Melke's outreachy internship) had a discussion on how this could help us to monitor the diversity in the Galaxy community.

Could we add more metadata to be able to monitor the diversity of the Galaxy Community and how we are (hopefully) making progress?
Information (of course not mandatory) such as:
- Gender
- Pronouns
- Ethnicity
- Since when in the community
- Active: yes/no  (for instance if you no longer work for the Galaxy community but still to have your individual page to reflect all the work you have done in the past; or when you are on leave, etc.).

Of course, we are happy to discuss it with you! Thanks.

NickSto:
Pronouns would be great to add, but it strikes me as somehow ill-advised to add a metadata field for ethnicity (and maybe even gender), even if it's optional. Theoretically it should be the same as if someone wants to describe themselves in the body of the page, but for some reason making it "data" feels uncomfortable? If we want to monitor diversity, I think something like a survey would be more standard. Of course, these are just white guy thoughts so I'm not the one to listen to on this.

As for the other fields, based on the pattern of existing personal pages, if we had a binary `active` field, it would always be `true`, since no one updates their pages after they leave. I'd suggest instead (or in addition) an `active_as_of` field with a date value so you know how out of date this activity assertion was. I guess the rare, super-responsible person could still set `active` to `false` after they leave, but I'm guessing the `active_as_of` will be more useful.

annefou:
You have very good points and I like your concrete and practical suggestions. Monitoring diversity is difficult and I agree that we should be extra careful. Thanks for your comments and suggestions: we need thoughts from everyone and we need to make sure everyone feels empowered to openly discuss these difficult topics. This is how we can grow together and build a more open and diverse community.

content/people/assunta/index.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Assunta DeSanto
+subsites: [us]
+groups: [wg-goats]
+github: assuntad23
+image: adesanto_book_resize.png
```
hexylena, line 6: could use an alt-text here for folks to self-describe their image.

Issue #1442, opened 2022-06-09
Reimplement JSON-LD data for events


NickSto:
This was implemented in Hub 1.0, and instead of reimplementing it after the rewrite, we went with the [Hub unification](/galaxyproject/galaxy-hub/projects/3) effort as a solution for crossposting content across communities. But:

> So a lot of european projects e.g. TeSS require json-ld, and it's incredibly useful for that
> they are the "training materials and event scraper for europe" from elixir
> removing this will break that ingest, unfortunately
> If it would be possible to resurrect that, it'd be really useful for european collaboration efforts!

\- @hexylena

Hub 1.0 had JSON-LD on the [/events/ page](http://web.archive.org/web/20210324100645/https://galaxyproject.org/events/): one blob per event, each in its separate `<script>` tag:

```html
<script type="application/ld+json">{ 
 "@context": "http://schema.org/",
   "@type": "Event",
   "name": "Speed up your data analysis with Galaxy: Features no one knows about","url": "https://galaxyproject.eu/event/2021-01-15-webinars-adv-features-session4/", "duration": "P1D",
"location": {
   "@type": "Place",
   "address": "Galaxy-ELIXIR Advanced Features webinar series, Online, Europe", 
"url": "https://us02web.zoom.us/webinar/register/WN_oIXQYaCpTfqp9RgOu6gj3g" },
"contact": {
 "@type": "person",
 "name": "Jennifer Hillman-Jackson, Anika Erxleben"
},"startDate": "2021-03-24"
}</script>
```
But @hexylena had some suggestions for improving that:
- some are Event, some are Course and they're semantically separate
- contact is a Many field, so we don't have to squash multiple people in as if they were one person

For reference, the original issue for adding it to Hub 1.0: #252

hexylena:
Thanks @NickSto! if y'all get something started, I can maybe work on making it match all the schema properties?

NickSto:
One question about implementation is whether to put json-ld blobs on the /events/ page and/or one blob on each individual event's page. The old .org Hub did the former and .eu does the latter.

According to @hexylena, if we have a sitemap, sites like [TeSS](https://tess.elixir-europe.org) can use that to find every event page and extract the json-ld from there, so we don't have to include the blobs on the /events/ page.

We don't currently have a sitemap, but there's a [Gridsome plugin](https://gridsome.org/plugins/@gridsome/plugin-sitemap) for that, so hopefully it should be easy.

[...]

PR #1564, opened 2022-07-29
Add some JSON-LD data to pages

NickSto:
This is the start of an implementation of #1442.

I followed roughly the example given in that issue (which is what the old Hub produced). Schema.org doesn't show `contact` as a valid property of `Event`, but the old Hub did that and so does [SIB](https://www.sib.swiss/training/upcoming-training-courses), which TeSS gives as an example implementation. So I included it, if only to give an example of how to include contacts. One question about that is what to do when there's multiple people listed as a contact?

### DIFF
--- a/src/components/ArticleHeader.vue
+++ b/src/components/ArticleHeader.vue
@@ -46,7 +46,7 @@
 
 <script>
 import Redirect from "@/components/Redirect";
-import { ensureDomain, humanDateSpan } from "~/lib/utils.js";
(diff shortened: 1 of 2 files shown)


bgruening:
We have this now in the new Hub.

https://validator.schema.org/#url=https%3A%2F%2Fpreview.galaxyproject.org%2Fnews%2F2025-12-12-galaxy-newsletter%2F

PR #1568, opened 2022-08-03
Redesign Homecards: light theme

NickSto:
This gives the homecards (the panels on the front page) an overdue due redesign.

I basically ripped off the look of the cards on the .eu homepage because they look way better.

Feedback is welcome!

image

### DIFF
--- a/src/components/HomeCard.vue
+++ b/src/components/HomeCard.vue
@@ -55,7 +55,6 @@ export default {
     margin-right: 0.4em;
 }
 .content {
-    border-top: 1px solid #ddd;
     padding: 12px;
 }
 .content.markdown::v-deep > *:last-child {
--- a/src/components/HomeCard.vue
+++ b/src/components/HomeCard.vue
@@ -8,7 +8,7 @@
                 </g-link>
             </h2>
             <div class="body">
-                <div class="markdown content" v-if="hasContent(content)" v-html="content" />
+                <div class="markdown content" v-if="content && content.trim()" v-html="content" />
                 <ItemListBrief v-for="(item, i) in items" :key="item.id || i" :item="item" :target="target" />
             </div>
         </div>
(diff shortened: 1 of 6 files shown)


beatrizserrano (approved):
Looks good to me, Nick! This style definitely fits better, thanks. 
The only thing that bothers me a bit is the different widths. I think that what fails is the cards with the profiles, I don't know how to make them wider.
image

NickSto:
@beatrizserrano Okay, I've tweaked the cards so their row is *nearly* the same width as the columns below. It's tricky to get it exact, but maybe this is close enough?

image

beatrizserrano:
> @beatrizserrano Okay, I've tweaked the cards so their row is *nearly* the same width as the columns below. It's tricky to get it exact, but maybe this is close enough?

Yes :) Thanks for finding a solution!

dannon:
I'm tinkering with a few options here and I think it's easier to read and less cluttered if we just drop the borders everywhere.  What do folks think?

Before:
image


After:
image

NickSto:
@dannon Hmm that's a toughie. I tend to like the defined panels, but I'm no designer so I wouldn't weight that opinion heavily.

beatrizserrano:
Maybe just the outer border would work? i.e. separation between boxes but not between items. That would be an intermediate way of presenting it

dannon:
@beatrizserrano Sure, let me mock that one up

NickSto:
Still hard to tell for me. I think I like this more than no borders at all, personally.

beatrizserrano:
Yes, I like it more, too! Thanks, @dannon and @NickSto :)

dannon:
@beatrizserrano Okay, cool.  Let me push that to this branch and then we can get this in!  We can always continue to iterate, of course.

NickSto:
Looks good to me. I'm ready to merge.

NickSto:
Man, it feels so much better to go to the homepage and see this instead now.

Issue #1569, opened 2022-08-03
What to do with .eu carousel content?


NickSto:
Most of the static content on the .eu homepages and [middle panes](1380) is in their carousels. As we migrate these pages to .org, we need to figure out what to do with this content. No one's keen on carousels anymore, so we'll be putting in regular page sections. And it sounds like a lot of the content is legacy or unneeded, so we won't be keeping all of it.

As a first step, I'm charting out what info each carousel holds. A lot of it is common content.

| Item | [EU home](https://galaxyproject.eu) | [EU middle](https://usegalaxy-eu.github.io/galaxy/) | [.org/eu/](https://galaxyproject.org/eu/) | [Frei home](https://galaxyproject.eu/freiburg/) | [Frei middle](https://galaxyproject.eu/freiburg/galaxy/) | [Eras home](https://galaxyproject.eu/erasmusmc/) | [Gen home](https://galaxyproject.eu/genouest/) | [Lani home](https://galaxyproject.eu/elixir-it/) | [IFB home](https://galaxyproject.eu/ifb/) |
|-----------------------------------|:---:|:---:|:----:|:---:|:---:|:---:|:---:|:---:|:---:|
| Data Driven Research for Everyone |  ✓  |  ✓  | text |  ✓  |  ✓  |  ✓  |  ✓  |     |  ✓  |
| about UseGalaxy.eu                |  ✓  |  ✓  |  ~   |     |  ✓  |     |     |     |     |
| about the instance                |  ✓  |  ✓  |  ~   |     |     |     |     |  ✓  |     |
| (quota increase link)             |  ✓  |     |  ✓   |     |  ✓  |     |     |     |     |
| Domain-specific Galaxies          |  ✓  |  ✓  |      |     |     |     |     |     |     |
| TIaaS                             |  ✓  |  ✓  | link |     |  ✓  |     |     |     |     |
| Training                          |  ✓  |  ✓  |      |  ✓  |  ✓  |  ✓  |  ✓  |     |  ✓  |
| the team what run it              |  ✓  |  ✓  | text |     |     |  ✓  |  ✓  |  ✓  |  ✓  |
| Downtime Notifications            |  ✓  |  ✓  |      |     |     |     |     |     |     |
| Data Retention Policy             |  ✓  |     | link |     |  ✓  |     |     |     |     |
| Maintained/funded by              |  ✓  |  ✓  |  ~   |     |  ✓  |     |     |     |     |
[...]

PR #1573, opened 2022-08-09
Add content to EU subsite homepages and middle panes

NickSto:
Currently, the homepages for the EU subsites (aside from `eu` itself) all have the same content: basically a generic boilerplate placeholder. So this PR is an effort to differentiate and customize them, using the existing content from galaxyproject.eu.

Since the homepages share a lot of content between them, part of this PR is a system to make it easy to plug in existing Inserts into homepages by just adding a symlink. Basically, the current homepages have a handful of slots you can put an Insert into, like `main.md`, `lead.md`, and `lower.md`. This PR modifies that so you can have multiple Inserts per slot, e.g. `lead1.md`, `lead2.md`, etc. These can be symlinks to existing Insert files, so you don't have to have duplicate content all over the place.

This also takes the opportunity to implement the middle panes for each of the subsites (described [here](/galaxyproject/galaxy-hub/issues/1380#issuecomment-1171720481)), since they use a lot of the same content.

This presents one answer to #1569: just keep all the carousel content for now. We can always change or remove things later.


beatrizserrano (approved):
Nice @NickSto :) Thank you! I can't comment much on the content, but will let the partners know that they can update as they like, or even create new subsites!

NickSto:
I'll go ahead and merge for now, especially since they're not yet public. We can iterate if we'd like.

PR #1577, opened 2022-08-11
Rename "Support" NavBar dropdown to "Help".

NickSto:
I think "Help" is a better description of what you'll find under the current "Support" dropdown. To me, "Support" suggests more like live software support, but the things in the dropdown are more like self-help resources.

It also gives us just a smidge more room in the navbar, which really matters at this point.

Before:
image

After:
image

### DIFF
--- a/navbars.json
+++ b/navbars.json
@@ -6,7 +6,7 @@
         ],
         "right": [
             {
-                "label": "Support",
+                "label": "Help",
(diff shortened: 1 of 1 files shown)


dannon (approved):

beatrizserrano (approved):
makes sense to me too!

PR #1582, opened 2022-08-12
First draft of custom navbar for each subsite

NickSto:
This is my attempt at giving each subsite a custom navbar based on the ones they have at galaxyproject.eu. Ideally I'd like to have put everything they have in their .eu navbars on the left-hand "subsite-specific" section, and keep the right side constant site-wide.  But there was usually too much to fit it all, so in several cases I put some of the links in dropdown menus on the right side. Sometimes I instead removed items from the right side to fit them. In all cases, every link in the .eu navbar was kept somewhere, whether in a dropdown or its own top-level item.

It was a lot of judgment calls on what things people would like to give top billing over others. I don't know what the priorities are for the different communities, so I could use feedback. @beatrizserrano @bgruening

image
image
image
image
image
image
image
image


beatrizserrano:
Hard for me to judge what's best as well. We can leave it like this for now and ask the member sites what they think about it.

PR #1631, opened 2022-09-14
Rework how Bundles are gathered, don't include HomeTop when empty.

NickSto:
This reworks `gatherBundles()` and omits the HomeTop component when it's empty. This has the effect of, for instance, removing the excessive empty space below the title on the Pasteur homepage.


NickSto:
At this point it's just the Gitter test that's failing (#1632) so I'll go ahead and merge.

PR #1636, opened 2022-09-20
Swap community call weeks Oct27 and Nov17.

dannon:
Was suggested to get the release conversation out there a bit earlier before the 23.1 freeze.  Cleared with @hexylena.


beatrizserrano (approved):
good idea :)

Issue #1317, opened 2022-04-22
How should posts be aggregated across subsites?


NickSto:
In the Hub meeting this morning we confronted an issue central to the Hub unification effort.

One of the main motivations of unification was making it easy to cross-post content across different Galaxy locations. Instead of having to create a post on .org that points to a .eu event, you can just tag the EU event with `us` and it will show up in the US feed as well.

But then should that post show up in the global, Galaxy-wide feed? What about an event in an EU subsite like Pasteur? Should all Pasteur events be listed in the EU-wide feed? In the global feed? What about aggregation in the other direction? Should the GenOuest feed list all EU events? All global events?

Theoretically it seems like the answers to all these questions should be yes: federation everywhere! Why not make everyone aware of everything? But I can imagine instances where you'd like to be able to post something to only one subsite (or a few). What if there's an event which will only be in French? There are [a](https://github.com/usegalaxy-eu/website/blob/master/_posts/2021-01-18-usegalaxy-fr-hiring.md) [few](https://github.com/usegalaxy-eu/website/blob/master/_events/2021-05-03-agrocampus-1.md) [examples](https://github.com/usegalaxy-eu/website/blob/master/_events/2021-05-03-agrocampus-2.md) which were posted only to French sites like GenOuest, Pasteur, VIB, and ELIXIR-FR/IFB. I could also imagine subsites might want to be a place to see *only* the stuff happening in that particular area, without all the noise of all the general EU/Global stuff.

Currently it seems the .eu site only propagates up: a post will appear only on the member sites which were tagged, plus the EU-wide feeds. If we kept that model, then the global Galaxy news & events feeds would list every post from everywhere. This might be okay, or it might be too much noise. Keeping strictly to the current .eu model would also mean we'd have to specifically tag every single subsite if we wanted a post to show up everywhere, since there's no "higher-level" site tags on .eu (like `global`, `eu`, or `us`).

Probably the most helpful thing is looking at some examples. I started a [spreadsheet](https://docs.google.com/spreadsheets/d/1b-9qXES45klWO3e644Ugiadkfkhx1ul7GRLOUcqXSzI/edit?usp=sharing) where we can pick out existing posts and list where they currently are listed, and then say where we think they *should* be listed. Please give your input by classifying some posts.

bgruening:
The spreadsheet is a great idea, I filled it.

[...]

PR #1640, opened 2022-09-22
Add pages listing every news and event post from every subsite

NickSto:
With the decision in #1317 to make every subsite only list posts tagged specifically for it, no current page, even the `global` ones, contains a full list of every post. This could be useful for various reasons. Right now, it's very useful in the .eu content migration process to make sure we have every one of their posts.

The urls are `/everywhere/news/` and `/everywhere/events/`. I'd have preferred to use `all` instead of `everywhere`, but that could get confused with the `all` subsite shorthand, which whose definition is whatever `config.json` says, so it doesn't necessarily mean the same thing.


NickSto:
Seeing no objections here, I'm going ahead and merging this one. It seems useful enough.

PR #1679, opened 2022-10-17
Give more visibility to communities

beatrizserrano:
A few changes, following up on the discussion during the EGD CoFest with @bebatut @hrhotz @NickSto:
- Give a more prominent place to the regional communities and communities of practice (separated pages and link in masthead).
- Rename projects to applications. That way, the communities of practice will have a link to their own pages in the future. Those pages will gather the different projects in which that community is working on (@bebatut has plans for microGalaxy in this direction). Related to https://github.com/galaxyproject/galaxy-hub/issues/1428

Please feel free to edit if the changes don't match what we discussed! :)


NickSto:
Looks pretty cool to me.
One small suggestion: how about simpler urls, like `/community/practice/` and `/community/regional/`?

beatrizserrano:
> Looks pretty cool to me. One small suggestion: how about simpler urls, like `/community/practice/` and `/community/regional/`?

Great suggestion @NickSto, I've made the change already :)

PR #1692, opened 2022-10-25
List VueArticles in dynamic pages too

### LINKED ISSUE #751: List VueArticles in dynamic pages
Currently, dynamic pages (`/news/`, `/events/`, `/events/webinars/`, etc) won't list any pages that use Vue components in their Markdown (`VueArticle`s). This issue actually doesn't have an effect right now, since none of those pages use Vue. But a big motivation for the new Hub is to allow Vue to be used anywhere, so we don't want pages to be unexpectedly missing from these listings when an author uses a Vue component.

The issue is that in the dynamic pages, the GraphQL queries are for `allArticle`, not `allVueArticle`. The naive approach will be to add a corresponding `allVueArticle` query for each `allArticle` one, then somehow interleave them in the template.

Ideally we'd find ways to deduplicate the queries and abstract away the interleaving.

NickSto:
This resolves #751 by creating a new Collection called `ParentArticle`. Each `ParentArticle` is created by copying the data from an `Article` or `VueArticle`. But FYI the "copies" are mostly just references to the same string objects, so there's not much memory overhead.

So dynamic pages can just query for `allParentArticle` instead of doubling the number of queries (one for `Article`s and one for `VueArticle`s). Since the GraphQL queries are the least dynamic part of the system, without this there's not much we can do to avoid hardcoding two copies of every query.

This was the least bad solution I found. I also [tried](https://github.com/NickSto/galaxy-hub/commit/93ddb29d97993069334a9b1baff178be5d338c02) creating a sort of inheritance through GraphQL interfaces, but I don't think Gridsome offers enough access to the schema to make it work.

FYI, I've tested and this doesn't add any significant time or memory usage to the build (or development server).

### DIFF
--- a/gridsome.server.js
+++ b/gridsome.server.js
@@ -530,7 +530,7 @@ module.exports = function (api) {
     api.createPages(async ({ graphql }) => {
         eventsData = await graphql(`
             query {
-                allArticle(filter: { category: { eq: "events" } }) {
+                allParentArticle(filter: { category: { eq: "events" } }) {
                     totalCount
                     edges {
                         node {
@@ -588,7 +588,7 @@ module.exports = function (api) {
 
 function makeCalendar(eventsData) {
     let events = [];
-    for (let event of eventsData.data.allArticle.edges) {
+    for (let event of eventsData.data.allParentArticle.edges) {
         event = event.node;
         if (event.date) {
             const evt = {};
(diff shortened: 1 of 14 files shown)


dannon (approved):
Thanks for working on this; it looks like a fine workaround to me.  We can follow up on the one exception down the road and revisit it as necessary.

PR #1699, opened 2022-10-27
Fix issues with using images in /assets/media/

NickSto:
This adds a symlink from static/assets to content/assets so we can place images from .eu there and have them work with unchanged absolute urls like /media/assets/image.jpg. I also had to add /assets/ to the allowlist of urls which shouldn't be altered by fix-links.mjs.


bgruening:
Nice!

PR #1700, opened 2022-10-27
Add all .eu news posts since Oct 2021

NickSto:
I used the port-eu-post.py to import all news posts back until the beginning of October 2021. I did a cursory manual review each time, but there may still be things broken.


bgruening:
Thanks @NickSto 

Do we have a similar script for the events?

NickSto:
@bgruening The script will work for events as well. Those are next!

bgruening:
Awesome, thanks!

PR #1705, opened 2022-10-31
add initial Makefile with install and run targets

bgruening:
Given a conda environment a simple:

```
make install
make run
```

will get you started. The ugly openssl hack that I needed is included.


dannon:
I'm actually not sure if more folks will come to this repo with a functional `make` and related tools, or the javascript tooling of `node` and `yarn`, which we've tended to guide folks to, but this can't hurt to add as an option.  Thanks!

PR #1710, opened 2022-11-01
Remove migration notices from subsite news and events pages

NickSto:
With #1707 merged, the migration notices are no longer correct. We *do* have all the old news and events posts from .eu. They have missing, broken, or unreadable parts, but they're there.


bgruening (approved):
Very very cool! Thanks a bunch!

PR #1727, opened 2022-11-11
Add ESG page for preview purposes

ElectronicBlueberry:
Adds the Euro Science Gateway project page, so it can be previewed and shared. This page is still a WiP, and not linked in the navbar.

image


src/pages/projects/esg/Index.vue
```diff
@@ -0,0 +1,358 @@
...
+                <h2 class="card-header mt-0">
+                    <span class="icon fas fa-bullhorn mr-1" />
+                    News
+                </h2>
+                <div class="card-body p-2">
+                    <div>
+                        <ItemListBrief v-for="edge in $page.news.edges" :key="edge.id" :item="edge.node" />
+                    </div>
+                    <a :href="esgUrl('/projects/esg/news')" class="btn btn-secondary border-0 btn-sm w-75">
+                        <b>More News</b>
```
dannon, line 73: I can't decide if I like this button more than the other approach in the rest of the hub where the header is the link, but we should probably pick one and standardize for these itemlist components.

ElectronicBlueberry:
I'm noticing two issues which didn't occur locally. The solo view does not work, (https://galaxyproject.org/projects/esg?solo=true) and on the esg page the navbar dosen't work. I'll need to look into those.

dannon:
> I'm noticing two issues which didn't occur locally. The solo view does not work, (https://galaxyproject.org/projects/esg?solo=true) and on the esg page the navbar dosen't work. I'll need to look into those.
> 
> Edit: all local links seem to be non-functional, not just the navbar ones

This almost always means malformed tags somewhere in the content on the page.

PR #1730, opened 2022-11-14
Several fixes for ESG

ElectronicBlueberry:
Followup for #1727
Fixes malformed tags and the solo view


bgruening (approved):
Thanks @ElectronicBlueberry

Issue #1731, opened 2022-11-15
Search parameters are redirected


ElectronicBlueberry:
This link should lead to an alternate view of the ESG page: https://galaxyproject.org/projects/esg?solo=true

Instead it is redirected to the default view. Upon further inspection, all URLs containing search parameters are redirected with a 302 response. I can not reproduce this behavior locally, even with a production build, so I suspect this may have to do with the nginx configuration.

dannon:
In some quick testing it looks like it's the lack of a trailing slash causing the issue; https://galaxyproject.org/projects/esg/?solo=true works fine.  My guess is nginx trying to append slashes to the end of urls.

bgruening:
Who is in charge of the nginx? Is that @natefoo?

ElectronicBlueberry:
I could try changing all the internal links to contain a trailing slash

dannon:
Yeah, using slashes as appropriate on the URLs is a fine quickfix.

Here's the configuration for galaxyproject.org if you want to take a look:
https://github.com/galaxyproject/infrastructure-playbook/blob/main/env/galaxy/templates/nginx/galaxyproject.j2

natefoo:
It seems odd we should need to do the redirect to append a `/` in the proxy - I've never seen an application not handle this correctly.

dannon:
Is the 302 actually coming from aws?

![image](https://user-images.githubusercontent.com/155398/202259371-af1f807a-e185-4572-88fa-133a3a738971.png)

natefoo:
Good catch, yes, this is indeed S3. Specifically I believe it's the automatic index.html feature for static website serving. Here is a similar discussion: https://github.com/jariz/gatsby-plugin-s3/issues/51

Is there a reason we didn't just make this served by CloudFront? I am not sure if it would have the same issue (but could be addressed with the Lambda@Edge rule there, but in general it'd probably be preferable to being served from an EC2 instance.

dannon:
No specific reason for picking this model; I think that's just what we already had set up for something else?  

I originally really wanted to just use netlify but the build is too big for my free plan; I can revisit options.  There was a cloudflare version I wanted to try, too.

natefoo:
Ok, just let me know. I have 4 different sites using CloudFront now so this should be relatively straightforward, except maybe for getting certs on the root of a domain not in Route 53.

dannon:
[...]

PR #1732, opened 2022-11-15
ESG website - followup 1

ElectronicBlueberry:
Fixes all points from https://github.com/usegalaxy-eu/issues/issues/349

Additionally adds link icons to all headings.


src/pages/projects/esg/Index.vue
```diff
@@ -144,6 +157,9 @@ export default {
                 return url;
             }
         },
+        slugify(string) {
```
dannon, line 160: We have `@sindresorhus/slugify` as a depenency already for other path manipulations if you just want to reuse that here.

dannon (approved):

bgruening:
Cool thanks a lot.

ElectronicBlueberry:
closes #1734
may fix #1731

content/projects/esg/main.md
```diff
@@ -3,9 +3,7 @@ title: EuroScienceGateway
 description: Improving computational workflows in Galaxy by maturing Pulsar Network and publishing computational workflows as FAIR Digital Objects.
 ---
 
-# EuroScienceGateway Project
-
-![eosc | EuroScienceGateway](./eosc_euro_science_gateway.svg)
+![eosc | Euro Science Gateway](./eosc_euro_science_gateway.svg)
```
stain: Seem to be reverting #1733, change back to EuroScienceGateway

src/layouts/ESG.vue
```diff
@@ -30,6 +30,11 @@
 
         <main id="maincontainer" class="container">
             <slot />
+            
+            <div class="funding mt-2">
+                <g-image class="eu-img" src="/images/esg/EN-Funded by the EU-POS.png" width="200" />
```
stain, line 35: Don't use spaces in image filename as you need to `%20` escape them.

stain (changes requested):
Great! Just need to fix the image filenames to not have spaces (or use `%20` escapes), and also remove spaces from "Euro Science Gateway" alt text.

content/projects/esg/main.md
stain: I would also keep the `# h1` header as otherwise there is not textual header on page.

content/projects/esg/main.md
```diff
@@ -3,8 +3,6 @@ title: EuroScienceGateway
 description: Improving computational workflows in Galaxy by maturing Pulsar Network and publishing computational workflows as FAIR Digital Objects.
 ---
 
-# EuroScienceGateway Project
```
stain, line 6: Don't remove the header!

bgruening: The idea was to remove the heading so ce the big logo is underneath with the same text... Should we keep the header ?

bgruening: Stian if you think we should bring it back, let us know and we will do so.

PR #1747, opened 2022-12-01
Fix esg wp links 

dannon:
using lowercase paths will prevent the slug algorithm from trying to break these apart


bgruening (approved):
Interesting! Thanks!

PR #1769, opened 2022-12-10
News entry for PR (13113): "Add Storage dashboard"

bgruening:
This is a test PR that takes a PR description and creates a small news entry for it. The second commit is what the reviewer of this hub PR needs to do manually. The bulk of technical descriptions and images should already be there so that the hub reviewer only needs to clean up and bit and maybe simplify the sentences.

The date is taken from the pr-merge-date, so we can also create these posts retrospectively. There are some really good PR descriptions with (animated) images or even videos, I think they are good content for our general audience that we should re-use in the hub.

Any feedback and comments?

### DIFF
--- a/content/news/2022-03-25_add-storage-dashboard/index.md
+++ b/content/news/2022-03-25_add-storage-dashboard/index.md
@@ -1,5 +1,5 @@
 ---
-title: Add Storage dashboard
+title: Add Storage dashboard from David López
 date: "2022-03-25"
 tease: "New pull request from David López merged: Add Storage dashboard"
 hide_tease: true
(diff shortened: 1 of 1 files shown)


bgruening:
grafik

content/news/2022-03-25_add-storage-dashboard/index.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Add Storage dashboard from David López
+date: "2022-03-25"
+tease: "New pull request from David López merged: Add Storage dashboard"
+hide_tease: true
+autotoc: false
+tags: ['area/UI-UX', 'kind/feature', 'highlight']
```
bgruening, line 7: @dannon do you think we can just stick to the Galaxy tags here?

ElectronicBlueberry: I think using different tags based on the galaxy tags may be better here. I'd remove the `area/` and `kind/` prefixes, as these seem more like technical details, than what users would search for.

bgruening: Have done so, thanks Laila.

PR #1771, opened 2022-12-12
Update regional communities links

wm75:


wm75:
External links are now consistently using the corresponding icon,
internal links use hyperlinked text. Links for .eu are now more
consistent with the ones on its dedicated page, and include a matrix link.

dannon (approved):

PR #1773, opened 2022-12-12
Snowflakes for the holiday season

dannon:
Just an example; we can prune this back and only apply to a subset of pages if desired.

Can also add a button to disable them if we want, I guess.

If we want this on usegalaxy welcomes, I can add it there, too.  Then the instances get the snowflake treatment as well, automatically, without having to deploy anything specific to them.

https://user-images.githubusercontent.com/155398/207084535-979d6385-3151-4a3a-a220-6dd452215c6b.mp4

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,7 @@
 {
     "host": "galaxyproject.org",
     "contentDir": "content",
+    "snowflakes": true,
     "collections": {
         "BareArticle": { "path": "/bare/", "type": "vue" },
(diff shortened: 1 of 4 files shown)


bgruening (commented):
Cool, I like it :)

assuntad23 (approved):

bgruening:
So many hearts and many :+1: ... merged :) Happy holidays!

PR #1774, opened 2022-12-14
Refactor snowflakes to component, add to welcome,  add duration

dannon:
Defaults to 60 seconds on, then they go away.

Drops global config -- yagni.


bgruening:
merci, very cool idea!!!

PR #1776, opened 2022-12-14
No snowflakes if prefers-reduced-motion

### LINKED ISSUE #1775: respect prefers-reduced-motion with snowflakes
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

maybe just disable them if prefers-reduced-motion is set?

they're cute otherwise!

dannon:
fixes #1775

### DIFF
--- a/src/components/Snowflakes.vue
+++ b/src/components/Snowflakes.vue
@@ -21,13 +21,17 @@ export default {
         maxSize: { type: Number, required: false, default: 35 },
     },
     mounted() {
-        const snow = new Snowflakes({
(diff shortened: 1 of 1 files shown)


hexylena (approved):
looks good to me.

dannon:
Thanks.  And thanks for the suggestion -- it was a good idea.

hexylena:
sure thing! I have had it on since forever, since I find all animation really distracting, so I really notice when there's motion :) Thanks for the quick fix

PR #1777, opened 2022-12-15
Fix openssl issue for webpack4 and node18

dannon:
Webpack has stated they won't be fixing this in webpack4x, which we are stuck with w/ gridsome for now.  This workaround workaround works for me and allows building/serving/developing with node v18.

Edit: I need to look at what node version ubuntu-latest is using here and why it's yelling about that option being used, I haven't seen this in the Galaxy context before.  Might just be able to update base image and move on, but I'll follow up on this later (after the Galaxy freeze likely) -- no rush on the PR.

### DIFF
--- a/config.json
+++ b/config.json
@@ -29,14 +29,34 @@
     "subsites": {
         "shorthands": {
             "all": [
-                "freiburg", "pasteur", "belgium", "ifb", "genouest", "erasmusmc", "elixir-it", "au", "eu", "us", "global"
+                "freiburg",
+                "pasteur",
+                "belgium",
+                "ifb",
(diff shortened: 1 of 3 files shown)


wm75:
Can confirm that this is working for me with nodejs 18 on fedora.

bgruening:
> /usr/local/bin/node: --openssl-legacy-provider is not allowed in NODE_OPTIONS
error Command failed with exit code 9.

dannon:
Made this more robust similar to what I did in https://github.com/galaxyproject/galaxy/pull/15756/files; this should be compatible with node 16+ now.

PR #1786, opened 2022-12-23
Ghost sites

neoformit:
- [x] New feature - ghost sites (see below)
- [x] Add AU as ghost site
- [x] Add JSON feed `/events/feed.json`

image

### DIFF
--- a/config.json
+++ b/config.json
@@ -29,14 +29,14 @@
     "subsites": {
         "shorthands": {
             "all": [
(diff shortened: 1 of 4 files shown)


dannon:
@neoformit Thanks, this is a great way to link in the AU site and still share events!

neoformit:
Thanks @dannon I'm interested to see how it goes! Galaxy Media Site is already reading an external news feed so it should be fairly easy to add another.

PR #1808, opened 2023-01-11
Update events metadata info in CONTRIBUTING.md

nakucher:
Was using this as a reference and noticed some links and metadata tags were out of date. 

Needs review of `location` tage and to define `tags` and `subsites` tags.


bgruening (approved):
Thanks a lot @nakucher! This is much needed. :)

PR #1836, opened 2023-01-25
edits to VGP project page

abueg:
hello! 👋🏼 

some edits to the VGP project page -- so far I have looked over the workflow details and intro text, but will take a look at results either tonight or tomorrow 👍🏼
cc @nekrut


content/projects/vgp/workflows/index.md
```diff
@@ -3,23 +3,23 @@
...
... -0.34375q0.015625 -0.046875 0.046875 -0.125l-1.0625 -2.765625l0.515625 0l0.578125 1.609375q0.109375 0.296875 0.1875 0.640625q0.09375 -0.328125 0.203125 -0.640625l0.59375 -1.609375l0.46875 0l-1.0625 2.8125q-0.171875 0.453125 -0.265625 0.625q-0.125 0.234375 -0.28125 0.34375q-0.15625 0.109375 -0.390625 0.109375q-0.125 0 -0.296875 -0.0625zm2.6796875 -1.0625l0 -3.8125l0.46875 0l0 3.8125l-0.46875 0zm4.4609528 0l0 -0.34375q-0.265625 0.40625 -0.765625 0.40625q-0.328125 0 -0.609375 -0.171875q-0.28125 -0.1875 -0.4375 -0.515625q-0.140625 -0.328125 -0.140625 -0.75q0 -0.421875 0.125 -0.75q0.140625 -0.34375 0.421875 -0.515625q0.28125 -0.1875 0.609375 -0.1875q0.25 0 0.4375 0.109375q0.203125 0.109375 0.328125 0.265625l0 -1.359375l0.46875 0l0 3.8125l-0.4375 0zm-1.484375 -1.375q0 0.53125 0.21875 0.796875q0.234375 0.25 0.546875 0.25q0.296875 0 0.515625 -0.25q0.21875 -0.25 0.21875 -0.765625q0 -0.5625 -0.21875 -0.828125q-0.21875 -0.265625 -0.546875 -0.265625q-0.3125 0 -0.53125 0.265625q-0.203125 0.25 -0.203125 0.796875zm3.0882568 1.375l-0.4375 0l0 -3.8125l0.46875 0l0 1.359375q0.296875 -0.375 0.765625 -0.375q0.25 0 0.46875 0.109375q0.234375 0.09375 0.375 0.28125q0.15625 0.1875 0.234375 0.453125q0.09375
```
abueg, line 13: comments for the image: 
- the workflow & QC text for WFs 7,8,9 are black when they're white in the rest of the boxes
- for the trio inputs, I think I might prefer "Illumina (mat.)" and "Illumina (pat.)" instead of the sex symbols, to just convey that it's read as from different individuals (ie the parents)
- for the decision point of "purging needed?" I suggest just having "Purging</br>?" so it matches the rest of the decision point
- for WF7, the outputs will be hap1 assembly and hap2 assembly, not primary&alternate
- for WF7, possibly just have the outputs as "purged contigs" to make it agnostic of the type of haplotype that went into it (since we can feed primary, altnernate, hap1, mat, etc)
- for the inputs, I would put HiC as a required input data, as that's the scaffolding data we do always try to generate

abueg: oh gosh I didn't realize commenting on this line would put its entire length on the PR page, sorry

abueg:
I added some figure legend-like text for the results, it's currently in the alt. text section of the markdown but can be re-used to be put as a caption, or whatever

- line 197: solo data outputs, kmer profiling - I removed the transformed genomescope profile plot as we don't often have to use it
- general comment about the merqury plots: this is more a matter of preference, I personally prefer the `*.fl` plots from merqury, it's more useful to me when I can see the spectra overlapping instead of stacked, but again not an important thing to change! but either way it should be consistently using one type, I think -- for instance, the first merqury plots featured are a `fl` spectra-cn plot and a `st` spectra-asm plot
- maybe worth putting the genomescope profiles for the parents in the trio outputs?
- i'm a bit confused about which haplotype you're showing in the trio results -- for the contigs, the BUSCO has some missing genes, which is consistent with that being results from the paternal haplotype, because the BUSCO geneset has some sex-specific genes, so the paternal one would be missing the bulk of the Z-specific genes. but then after purging, the BUSCO results have fewer missing genes, which would be consistent if it were the maternal haplotype?

nekrut:
@abueg = thank for this. I will merge and then make further edits to address you comments

PR #1841, opened 2023-01-31
Add Smorgsboard 2023 event

bebatut:


hexylena (approved):
Thanks @bebatut

I really miss @beatrizserrano and Dave doing this, I never remember now to post to the hub.

PR #1856, opened 2023-02-07
Added an image to the CIRAD job posting.

tnabtaf:
Also set end dates (random ones) for 2 other jobs that are no longer active.


bgruening:
Thanks Dave!

tnabtaf:
Ja! I have now traded emails with Stéphanie Bocs across my last 3 jobs.  I quite like that feeling.

dannon (approved):

PR #1862, opened 2023-02-10
Update title and presenter

nakucher:


bgruening:
I took out the link we can add this later. Thanks @nakucher!

PR #1867, opened 2023-02-14
Update Galaxy Metabolomics Community Call

hechth:
The event doesn't show up on the website, maybe it is due to some problems in the md.


bgruening:
I was a bit faster, its now fixed I think.

PR #1871, opened 2023-02-21
Test fixing CI

dannon:
(maybe ubuntu-latest bumped default node version, causing this?)

Related to https://github.com/galaxyproject/galaxy-hub/pull/1777


dannon:
This does fix the build step; still need to follow up on why the redirects are causing the link checker to fail.

PR #1900, opened 2023-03-07
Basic eventzilla embed

dannon:
Adds an eventzilla embed component that takes an event id and injects everything it needs.

Shows up something like this:

image


afgane:
@dannon, we have an early version of the GCC schedule so it's getting to be time to merge this. I'm not confident in resolving the conflicts so could you please take a look?

dannon:
@afgane Resolved; it just needs to be included wherever you want it to show up like was in https://github.com/galaxyproject/galaxy-hub/pull/1900/commits/52389502d07a3aa990804c4ee7a2a4f948a74994, but with the right event id.

afgane:
Excellent! Thank you. Let me check with Jenn if she imported the schedule yet and I can then add the event ID.

dannon:
Sounds good.  I'll go ahead and merge the component in, and it can be added to a page whenever.

PR #1905, opened 2023-03-11
Add news item for History Advanced Search

ahmedhamidawan:
Add a news item outlining the work of the new history search in the history panel.


bgruening:
Just a naive question. Is `extension` really the thing that we would like to advertise? Isn't there also `datatype` as a synonym available?

ahmedhamidawan:
> Just a naive question. Is `extension` really the thing that we would like to advertise? Isn't there also `datatype` as a synonym available? 
> 

I can change the first column value to be datatype (and the actual filter will remain the same) like i did for database (actual filter is `genome_build`)

bgruening:
> > is `extension` really the thing that we would like to advertise?
> 
> I think so ? I would assume this is more obvious to users than datatype ?

I think "extension" is confusing. We tell people all day that Galaxy datasets have no "filename" and "file-extension" and they should add rich metadata during upload.
"extention" is also a very unspecific word on its own. "extension" to what?

bgruening:
`format` is also ok with me.

dannon:
@bgruening I was about to say it probably doesn't matter, but those are gigantic and we don't want artifacts that big in the repo at any point -- but I don't know if the bot will do rebases.  I'll look into it, but @ahmedhamidawan can you rebase in a downscaled version of those easily enough?

PR #1917, opened 2023-03-16
Twitter post

erxleben:


bgruening:
@erxleben you can embed tweets directly into the news. See this change here.

PR #1920, opened 2023-03-16
Remove trailing slash in Lobby gitter chat URL (currently leads to a 404)

sanjaysrikakulam:
Trailing slash in the URL is leading to a 404 error.


dannon:
@sanjaysrikakulam Thank you!

dannon (approved):

PR #1921, opened 2023-03-16
Fix image in new history search news item

ahmedhamidawan:
Previous image had a history with an unusual tag (`@sampleA1` instead of `#sampleA1`) :

| Before | After |
| ------ | ----- |
| image | image |


bgruening (approved):

dannon:
I would never have noticed this, good eye.

PR #1927, opened 2023-03-17
Fix link in new history search news item

ahmedhamidawan:
I mistakenly linked `http://localhost:8080/news/2022-08-galaxy-release-22-05/` instead of `http://galaxyproject/news/2022-08-galaxy-release-22-05/`


dannon:
D'oh, and of course I didn't catch it looking locally

PR #1955, opened 2023-04-04
Slide rotation for usegalaxy.org welcome page

natefoo:
This is a component but the slides are hardcoded in it - we can parameterize it down the line but need to get something up quickly, and Dannon wanted to redo the page anyway since it's non-standard.

This was mostly me trying things and Dannon telling me what I was doing wrong, so big thanks to him for the help.


bgruening:
@dannon what do I need to do to put that on EU as well?

PR #1969, opened 2023-04-21
Act dont criminalize

wm75:
@bgruening I've included you as an author so make you agree with the wording, and ask for changes where you don't :)


bgruening:
@wm75 with the new website we need to update the notice here: `content/bare/eu/usegalaxy/notices.md`

bgruening (approved):
Thanks @wm75! Let's make some fuzz about it.

wm75:
> @wm75 with the new website we need to update the notice here: `content/bare/eu/usegalaxy/notices.md`

Ah, I see! Will do.

wm75:
Hope it'll work like this.

PR #1975, opened 2023-04-27
Add Galaxy devops position at Elixir-BE

pauldg:


bgruening:
@pauldg it needs a closing date.

PR #1987, opened 2023-05-04
Adding single cell community of practice

nomadscientist:
I haven't linked to a single cell instance because there are two (at least) and I don't want to play favourites!


nomadscientist:
Yes ma'am! Done

PR #1931, opened 2023-03-21
Three paper news

sebastian-schaaf:
First attempt of contribution to the news section of Galaxy Hub. Hope I met all rules and mechanisms.
Papers added/referenced:

- Rapp et al. 2023 - STAT3 signaling induced by the IL-6 family of cytokines modulates angiogenesis; J Cell Sci (2023) 136 (1): jcs260182; doi: 10.1242/jcs.260182

- Akol et al. 2023 - Multimodal epigenetic changes and altered NEUROD1 chromatin binding in the mouse hippocampus underlie FOXG1 syndrome; Proc Natl Acad Sci U S A. 2023 Jan 10;120(2):e2122467120. doi: 10.1073/pnas.2122467120

- Bray et al. 2023 - The Planemo toolkit for developing, deploying, and executing scientific data analyses in Galaxy and beyond; Genome Res. 2023. 33: 261-268; doi:10.1101/gr.276963.122

Rapp and Akol were carried out using galaxy.eu, Bray is a Galaxy community development.

PR #1990, opened 2023-05-05
Created news article for Paper Akol et al. on NEUROD1 chromatin binding

sebastian-schaaf:
One out of initially three submitted news articles out of PR #1931 (which thus could be rejected/deleted)


bgruening (approved):

bgruening:
Thanks!

PR #1992, opened 2023-05-05
Created news article for ESG/OpenAIRE MoU

sebastian-schaaf:
Finally... Hope it fits, and that it is ok to use the image that OpenAIRE created on their side...


bgruening (approved):
Thanks @sebastian-schaaf!

PR #2007, opened 2023-05-11
fix gcc2023 index button link for program

fubar2:
Fix program button link


afgane:
Thanks @fubar2 for noticing & fixing this.

fubar2:
It was Christina Hall who found the breakage - I assumed you were asleep :)

PR #2027, opened 2023-06-01
Add blog post on job resource selectors

kysrpex:
Just a small blog post to disseminate the _job resource selectors_ feature: menus that let the user configure the resources that the job may take advantage of.

<div style="text-align: center;">

</div>


kysrpex:
@anuprulez Can you please have a look at the blog post and maybe give some input if adequate?

content/news/2023-06-02-job-resource-selector-gpu/index.md
```diff
@@ -0,0 +1,29 @@
...
...or the GPU use-case you can take inspiration from [UseGalaxy.eu](https://usegalaxy.eu/)'s implementation: check out our [job_resource_params_conf.xml](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/files/galaxy/config/job_resource_params_conf.xml#L3-L6), [job configuration file](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/templates/galaxy/config/job_conf.yml#L465-L474), [list of configuration files](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/group_vars/sn06.yml#L516-L517), [Galaxy configuration file](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/group_vars/gxconfig.yml#L2180-L2186), [TPV tools configuration file](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/files/galaxy/tpv/tools.yml#L4-L12) and (within the latter file) [the configuration to be applied to each tool](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/06823c5fed9a4e589b02d46d3a224baa78596929/files/galaxy/tpv/tools.yml#L63-L64).
```
mvdbeek, line 29: I feel very ambiguous about this. The job resource selector implementation is quite bad, and using the job resource selector is very bad for portability of workflows and the overall UX ... why should a user decide if a job runs on GPU vs CPU ? why not try to make a better guess without involving the user ? AFAIK this was implemented to be able to test different job destinations, we should avoid this at all cost.

kysrpex: > using the job resource selector is very bad for portability of workflows and the overall UX

This is a quite good point, and I also agree that making a good guess is better that involving the user. 

@bgruening I assume the main benefit for us is that we can keep the load of GPU nodes (not so abundant in our cloud) low right? Since we do not have a guessing mechanism in place right now, our options are:
- Let the user decide (current solution)
- Send all tools that can use a GPU to a GPU node even if not needed for most datasets.
- Always use CPU. 

Maybe we should then not advertise how to implement it?

mvdbeek: I think you could also try something like https://total-perspective-vortex.readthedocs.io/en/latest/topics/tpv_by_example.html#metascheduling to send some jobs to GPU nodes if they are free.

bgruening: I don't like resource selectors at all and we should avoid them as much as we can. However, there are many people (especially in non-life-science communities) that want to use them - and are amazed if they find this feature.
I guess this is very community dependent and I don't want to judge them. Part of why we offer that is to have it as show case for those communities.

What about we add a text discussion on this as part of the blog post ... this feature exist, and we recommend using implicit rules vie TPV, but if you have the need for your community to make it explicit you can ...

content/news/2023-06-02-job-resource-selector-gpu/index.md
```diff
@@ -0,0 +1,29 @@
...
+tags: [EU]
+subsites: [all-eu]
+main_subsite: eu
+---
+
+The European Galaxy server relies on the compute nodes provided by the [de.NBI-Cloud](https://www.denbi.de/cloud). Did you know that some of them are **GPU-enabled**?
+
+Any Galaxy tools that require a GPU to function (e.g. [instaGRAAL](https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fbgruening%2Finstagraal%2Finstagraal%2F0.1.6%2Bgalaxy0&version=0.1.6%20galaxy0)) will automatically be scheduled to run on a GPU node by our [job router](https://galaxyproject.org/news/2023-05-08-tpv-switch/). However, there are tools that _can_ take advantage of a GPU, although they run on a CPU by default. Galaxy lets you select the type of compute node where such tools should be sent.
+
+For example, among the tools that can take advantage of a GPU is [Keras](https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fbgruening%2Fkeras_train_and_eval%2Fkeras_train_and_eval%2F1.0.8.4&version=1.0.8.4). On this tool's page, a _job resource selector_ that enables the job to be sent to a GPU compute node is displayed.
```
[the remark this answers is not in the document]
kysrpex: I asked you for a second review on your part. I will not ask you again regarding the [other discussion](https://github.com/galaxyproject/galaxy-hub/pull/2027#discussion_r1213423933).

mvdbeek (approved):
Thanks for the changes!

bgruening:
Thanks. Its also just displayed in EU sites.

PR #2064, opened 2023-07-05
Move "distancing badge" section to conduct page

neoformit:
Requested by Gareth

### DIFF
--- a/content/events/gcc2023/conduct/index.md
+++ b/content/events/gcc2023/conduct/index.md
@@ -20,3 +20,8 @@ a refund at the discretion of the conference organizers.
 
 Details of the the Code of Conduct can be read in the [Galaxy community Code of
 Conduct](/community/coc/).
+
+**The conference registration desk will offer stickers for adding to name badges
+that indicate personal preferences around physical distance (Red: I'm keeping my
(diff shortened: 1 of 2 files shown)


hexylena (approved):
Thanks @GarethPrice-Aus for the clarity

GarethPrice-Aus:
Hi @hexylena, sorry I don't understand the " :( " comment.  This request was in response to your email dated Friday 30th June - "Will you be moving the lanyard information away from the covid page to another page more appropriate so it doesn't have that confusing "this is a covid prevention measure" feeling?"
Another communication will be issued regarding the stance of the organising committee on mask wearing at GCC2023

hexylena:
sounds good @GarethPrice-Aus !

hexylena:
and you're right @GarethPrice-Aus the emoticon wasn't necessary, I'm just one of those millenials that is very transparent about their emotional state. Simply an expression of my feelings over the past five weeks since we began the topic, but you're right not relevant to the PR at hand. I'll remove it accordingly

GarethPrice-Aus:
I'll do the same and we can just re-write this bit of history

PR #2065, opened 2023-07-05
Automatically publish posts from GTN

### LINKED ISSUE #1417: Automatically pull in GTN posts
[update_gtn_posts.py](https://github.com/usegalaxy-eu/website/blob/master/update_gtn_posts.py) automatically pulls content from the GTN's [RSS feed](https://training.galaxyproject.org/training-material/feed.xml) (which tracks their [news posts](https://training.galaxyproject.org/training-material/news.html)) and adds them as new Hub posts.

We could use Gridsome's ability to pull in remote content and add it to the GraphQL database on build.

This might not be a high priority, though. It seems this hasn't successfully run since summer 2021.

kysrpex:
GitHub Actions workflow that automatically creates PRs to publish Galaxy Training Network news as external links.

Closes #1417.

Example PRs created by this workflow are available [here](https://github.com/kysrpex/galaxy-hub/pulls?q=is%3Apr+%22Publish+post%22+).

### DIFF
--- a/.github/workflows/gtn-news.yml
+++ b/.github/workflows/gtn-news.yml
@@ -138,14 +138,10 @@ jobs:
           """)["${{ matrix.name }}"]
           post = textwrap.dedent(f"""
               ---
-              site: [pasteur, freiburg, erasmusmc, elixir-it, belgium, genouest]
+              site: [all]
               date: "{post['published']}"
               tags: [training, gtn-news]
               title: "{post['title']}"
-              supporters:
-              - denbi
-              - elixir
-              - gallantries
               external: '{post["link"]}'
               ---
 
--- a/.github/workflows/gtn-news.yml
+++ b/.github/workflows/gtn-news.yml
@@ -1,0 +1,166 @@
+---
+# Pull news from the Galaxy Training Network's RSS feed and add them as
+# Galaxy Hub posts.
+name: Galaxy Training Network news
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 * * * *
+
+jobs:
+  collect:
+    name: Collect news from the Galaxy Training Network
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install feedparser.
+        run: pip install feedparser~=6.0
+
+      - name: Parse GTN RSS feed.
+        id: parse
+        shell: python -u {0}
+        run: |
+          import json
+          import os
+          from datetime import datetime, timedelta, timezone
+
+          import feedparser
+
+          feedparser.sanitizer._HTMLSanitizer.acceptable_elements.remove("a")
+          feedparser.sanitizer._HTMLSanitizer.acceptable_elements.remove("img")
+
+          feed = feedparser.parse("https://training.galaxyproject.org/training-material/feed.xml")
+
+          posts = {}
+          oldest = datetime.now().astimezone()
+          for entry in feed.entries:
+              feed_title = f"[GTN news] {entry['title']}"
+              feed_link = entry["link"]
+              feed_blurb = entry["summary"]
+              feed_published = entry["published"].split("T")[0]
+              feed_name = feed_link.split("/")[-1].replace(".html", "")
+              posts[f"{feed_published}-{feed_name}"] = {
+                  "title": feed_title,
+                  "link": feed_link,
+                  "blurb": feed_blurb,
+                  "published": feed_published,
+                  "name": feed_name,
+              }
+              feed_published_dt = datetime.fromisoformat(entry["published"])
+              oldest = feed_published_dt if feed_published_dt < oldest else oldest
+          oldest = (oldest - timedelta(days=1)).astimezone(timezone.utc).date()
+
+          with open(os.environ["GITHUB_OUTPUT"], 'w') as file:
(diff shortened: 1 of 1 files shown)


.github/workflows/gtn-news.yml
```diff
@@ -0,0 +1,166 @@
...
+              site: [pasteur, freiburg, erasmusmc, elixir-it, belgium, genouest]
+              date: "{post['published']}"
+              tags: [training, gtn-news]
+              title: "{post['title']}"
+              supporters:
+              - denbi
+              - elixir
+              - gallantries
+              external: '{post["link"]}'
+              ---
```
kysrpex, line 150: This needs to be double-checked.

hexylena: yeah not sure if we want to do something better with `supporters`, the GTN has a more extensive+accurate list, but keeping that in sync would be annoying and maybe not worth it.

Also the site selection, right now only EU ones but since it's now on the hub maybe it makes sense to distribute to all galaxies? (Or we can set a flag on the GTN side somehow to indicate that it's applicable for everyone?)

kysrpex: I better make this no longer a draft and request reviews from several people so that they also contribute to this discussion.

bgruening: Not sure the supporters will even be shown on the Hub with external links. I would either remove them completely or add something like "all" which we do not support at the moment.

`site:` I would definitely go for `all`.

.github/workflows/gtn-news.yml
```diff
@@ -0,0 +1,166 @@
...
+              site: [pasteur, freiburg, erasmusmc, elixir-it, belgium, genouest]
+              date: "{post['published']}"
+              tags: [training, gtn-news]
+              title: "{post['title']}"
+              supporters:
+              - denbi
+              - elixir
+              - gallantries
+              external: '{post["link"]}'
```
bgruening, line 149: My suggestion/vote would be:

```suggestion
              site: [all]
              date: "{post['published']}"
              tags: [training, gtn-news]
              title: "{post['title']}"
              external: '{post["link"]}'
```

hexylena: makes sense to me!

kysrpex:
I think we are ready to merge then, right?

.github/workflows/gtn-news.yml
```diff
@@ -0,0 +1,162 @@
...
+              feed_title = f"[GTN news] {entry['title']}"
+              feed_link = entry["link"]
+              feed_blurb = entry["summary"]
+              feed_published = entry["published"].split("T")[0]
+              feed_name = feed_link.split("/")[-1].replace(".html", "")
+              posts[f"{feed_published}-{feed_name}"] = {
+                  "title": feed_title,
+                  "link": feed_link,
+                  "blurb": feed_blurb,
+                  "published": feed_published,
```
hexylena, line 46: is there an author field on this end that takes plain text? our feed.xmk exposes it

.github/workflows/gtn-news.yml
```diff
@@ -0,0 +1,162 @@
...
+              for name in posts
+              if any(name in file for file in files)
+          }
+          for name in existing:
+              del posts[name]
+
+          with open(os.environ["GITHUB_OUTPUT"], 'w') as file:
+              print("posts=%s" % json.dumps(posts), file=file)
+
+      - name: Prepare matrix for `pull_requests` job.
```
hexylena, line 95: this seems complicated, why not have one PR that adds any (multiple) missing posts? it would be less work to manage than multiple PRs to approve and merge

.github/workflows/gtn-news.yml
kysrpex: If the feed exposes it then it is worth checking, I'll do it.

hexylena: cheers :) appreciate it!

.github/workflows/gtn-news.yml
dannon: PRs don't necessarily mean a lot of extra work, it's just more transparent vs committing to master.  Other than that these should be PRs, I don't have a preference over single or batched -- my guess is that only matters for the initial pass here and that they'll be largely single posts past that.

bgruening: I guess at least initially we should go for PRs for transparency. We can later auto-merge those after a given time or commit directly - which would save some CPU cycles and energy. But lets at first see if everything works.

Are we fine with merging as it is or doing the bulk-merge?
@dannon your call :)

hexylena: yeah PRs is good, just a single batch one would simplify this GHA *significantly* e.g. the [gtn's updater](https://github.com/galaxyproject/training-material/blob/2dd1a79f2e9578eee751c4a57f4616e772a0fbe6/.github/workflows/shortlinks.yaml#L43)

kysrpex:
I am very sorry that I closed the pull request, I should have made a new fork to test new changes on a clean environment rather than deleting the fork and re-creating it. Sorry again. I will open a new PR.

hexylena:
reopening a PR is also an option :)

kysrpex:
> reopening a PR is also an option :)

Doesn't work when you delete the repo :(

PR #2068, opened 2023-07-06
Automatically publish posts from GTN

kysrpex:
GitHub Actions workflow that automatically creates PRs to publish Galaxy Training Network news as external links.

Closes #1417. Folow-up of #2065.

An example PR created by this workflow is available [here](https://github.com/kysrpex/galaxy-hub/pull/21). An example workflow run is available [here](https://github.com/kysrpex/galaxy-hub/actions/runs/5477340935).

This PR hopefully addresses all the feedback from #2065, namely:
- Post template (@bgruening)
- Include authors (@hexylena)
- Batch all posts in one PR (@hexylena)

In addition, the whole approach has been reworked. The workflow uses now just feedparser and the GitHub REST API (through GitHub CLI and curl). Hopefully it is significantly faster (I cannot compare since I deleted the repo) as it avoids checking out the code.

PR #2070, opened 2023-07-07
Publish old news from the Galaxy Training Network

kysrpex:
Publish old news from the Galaxy Training Network that are not covered by the automation from #2068 [as requested](https://github.com/galaxyproject/galaxy-hub/pull/2069#issuecomment-1624905504).

Follow up of #2065.


bgruening (approved):
Very nice!

PR #2119, opened 2023-08-14
GCC2023 Meeting Report Blog Post - Fixed

natwhitaker:


bgruening:
@natwhitaker if you like we can help you review this in the beginning.

You submitted it as a blog post, but you probably wanted it as a news item? The difference between blog and news-item is not super clear and we discussed previously to remove blogs.

Do you need help to put the content of the PDF as news item?

PR #2120, opened 2023-08-15
Create 2023-09-small-scale event

hrhotz:


bgruening:
Do you maybe want to announce an in person bof at ESG here as well,? ,;-)

PR #2129, opened 2023-08-17
Create index.md for capquant

martenson:


nekrut:
@martenson add the following on top:

```
Histories for [A systems-level mass spectrometry-based technique for accurate and sensitive quantification of the RNA cap epitranscriptome](https://www.nature.com/articles/s41596-023-00857-0) by Wang et al. (2023).
```

PR #2148, opened 2023-08-28
Swap to blue logo as default opengraph/twitter include.

dannon:


dannon:
Better size, stands out better in previews.

PR #2161, opened 2023-09-11
Amend Galaxy EU 15.09.23 climate strike notification

kysrpex:
After looking at the outcome of #2159, I think it is very confusing that two banners have the same color and more importantly, that the strike banner has no title.

In addition, thinking it twice I think it makes better sense to link to the page specific to the action on Friday rather than the main page of Fridays for Future.


kysrpex:
@pavanvidem I cannot add you as a reviewer but I hope you agree too.

sanjaysrikakulam (approved):

PR #2168, opened 2023-09-13
Galaxy Newsletter September 2023 - Updated

natwhitaker:
Adding subsites
Checked and cleared all links


dannon:
Instead of a second PR with the same changes, I merged the original one and edited this so it's just the new link addition.

PR #2169, opened 2023-09-14
Fix meeting report location/link

dannon:


dannon:
Fixes the build after #2167, which caused an older news post to be navigable at which point the broken link got flagged.

PR #2192, opened 2023-09-28
Update Belgian Page

bedroesb:
Many things were out of date. This page should be more in line with the current Instance.


kysrpex (approved):
Your instance, your page, your rules :)

PR #2194, opened 2023-09-29
Egd2023 update schedule incl esg

sebastian-schaaf:
Several edits as discussed in the chat - still things to be done, but at least a step...


bgruening:
Thanks!

PR #2232, opened 2023-10-23
Drop Ukraine banner from usegalaxy.org welcome page

natefoo:


bgruening:
I added one commit to remove it from EU as well ...

PR #2257, opened 2023-11-14
Point to data source examples repo instead of gist

wm75:


wm75:
Redundant with https://github.com/galaxyproject/galaxy-hub/commit/72fc88b7436eb9082181fa935d49daea8a7d00e8.

PR #2296, opened 2023-12-20
Replace Gitter with Matrix

nsoranzo:


hexylena (approved):
Thanks for also standardising chat vs channel

PR #2304, opened 2023-12-26
Improved link to installation instructions for Galaxy-Hub

hujambo-dunia:


bgruening:
It's probably good to merge it, I'm just a bit concerned about the duplication of content. Do you think its worth the duplication?

hujambo-dunia:
@bgruening Thank you for this good question which is also true for the "big picture." However, this PR is more of a bug/content fix than it is adding new content. This PR is specifically to address users who cannot easily find the Developer Docs showing how to run Galaxy-Hub on a local environment. Myself including two others I spoke to (before reaching Dannon's helpful answer) did not know/recall how to run Galaxy-Hub locally. This PR makes those Developer Doc instructions a bit easier to find. There is [another landing page](https://galaxyproject.org/hub/) which is similar to this one, however, it does not mention local environment instructions (I'm glad to modify either page or both).

In regards to the "big picture", I agree with you that because of the novel way the Hub site was setup, the developer-friendly `README.md` Markdown file largely duplicates other parts of the user-friendly Vue/JS/CSS web pages. I personally found the existence of a `README.md` file in this case a bit confusing. So there is a larger question perhaps: (1) delete the `README.md` file altogether since the entire site is made up of Markdown files already (2) modify the `README.md` file so that it only contains links to the other Markdown files in the repo (without the duplicate content) or (3) keep the `README.md` file as it is - making edits where needed like in this PR? I'm happy to bring this up at the GOATS meeting on the 11th to start making inquiries to help solve the "big picture" issue.

bgruening:
Linking from the Readme.md to https://galaxyproject.org/hub/ seems good to me. The only downside I can see is for people working offline. But even in this case, you can not build the Hub because you can not fetch the dependencies.

hujambo-dunia:
> Linking from the Readme.md to https://galaxyproject.org/hub/ seems good to me. The only downside I can see is for people working offline. But even in this case, you can not build the Hub because you can not fetch the dependencies.

@bgruening Agreed - please see my most recent commits which allow for offline navigation from the `README.md` file.

content/hub/contributing/index.md
```diff
@@ -39,7 +41,9 @@ Once you're done, you'll want to add a "commit message" which describes your pro
 
 Then click "Propose new file" at the bottom. Then click the big, green "Create pull request" button. Then you'll see a couple more text boxes where you can describe your changes. This is the main post (called a "pull request") which the maintainers will read to understand what you're trying to do. You can use the commit message you already wrote if it's good enough for that. Once you're finished, click "Create pull request" at the bottom and the maintainers will look at it and decide if they want to add your changes. They might also reply with questions or requests for tweaks before they accept it.
 
-## Option 2: Clone the repository
+## Option 2
+
+## Developers: Clone the repository
```
hexylena, line 46: the empty section here is a bit odd (Especially for screenreaders.) maybe `Option 2: Developer: Clone the Repository` or `[Option 2] Deveopers: Clone the repo` or so?

hexylena (approved):
Looks like a nice re-organisation!

PR #2339, opened 2024-01-25
Travel updates 2

hechth:
Added more travel related information as well as activities during GCC.


afgane (approved):
Nice list of activities!

PR #2302, opened 2023-12-24
Start working on new governance pages

bgruening:
* introduction of the Community Board
* replacement of the Steering Board with the Community Board
* introduction of SIGs
* CoP are not grouped under SIGs
* regional communities are grouped under SIGs

* URLs are now looking like
  * /community/sig/genome-annotation/ or /community/sig/singlecell/
  * /community/sig/arabic/

grafik

## ToD:
* [ ] a new version of governance.svg
* [ ] Galaxy Technical Board (GTB) subpage (/community/governance/gtb) needs some content

PR #2341, opened 2024-01-25
Fix linkchecker report -- now it'll actually list broken links and their pages.

dannon:
It was always correctly reporting the number of broken links, but it was failing to correctly generate the helpful report of what those were.


dannon:
@bgruening This will help with your errors in https://github.com/galaxyproject/galaxy-hub/pull/2302

PR #2351, opened 2024-01-29
Home Page changes from GOATS-wg on 2024.01.23 (v2)

hujambo-dunia:
- Not yet ready for merging.
- Branch includes collective, but prioritized change requests from GOATS working group session on 2024.01.23

### DIFF
--- a/config.json
+++ b/config.json
@@ -124,5 +124,29 @@
         "/tutorials/rb_rnaseq/": "/tutorials/rb-rnaseq/",
         "/tutorials/var_dip/": "/tutorials/var-dip/",
         "/tutorials/var_hap/": "/tutorials/var-hap/"
-    }
+    },
+    "usegalaxy": [
+        {
+            "locale": "US",
+            "url": "http://www.usegalaxy.org",
+            "utcMin": -11,
+            "utcMax": -4
+        },
+        {
+            "locale": "EU",
+            "url": "http://usegalaxy.eu",
+            "utcMin": -4,
+            "utcMax": 7
+        },
+        {
+            "locale": "AU",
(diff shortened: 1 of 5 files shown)


config.json
```diff
@@ -124,5 +124,29 @@
...
+            "utcMax": 7
+        },
+        {
+            "locale": "AU",
+            "url": "http://usegalaxy.org.au",
+            "utcMin": 7,
+            "utcMax": 14
+        },
+        {
+            "locale": "All",
```
afgane, line 148:
```suggestion
            "locale": "Other",
```

content/home/education.yml
```diff
@@ -6,31 +6,39 @@ education:
 - title: Tutorials
   content: Access topic-based trainings for free.
   url: https://training.galaxyproject.org
+  icon: fa-solid fa-chalkboard-teacher
```
afgane, line 9:
```suggestion
  icon: fa-solid fa-graduation-cap
```

We use the graduation cap in the masthead already so probably best to stay consistent, despite my earlier comment in chat.

mvdbeek:
I don't think you necessarily need the tests to pass if you're not targeting the master branch. It looks like commits have just been pushed to the update-homepage-2 branch. I would however still try to figure out why this fails. It is possible that the base `update-homepage2` branch is already broken. As mentioned on gitter, https://github.com/galaxyproject/galaxy-hub/blob/master/.github/workflows/main.yml#L34-L39 is how you run the tests.

afgane:
The failing link checker tests are specific to the ESG project pages and it's not evident what's the issue with the GH Action failures so merging this as is and hoping the failures are an artifact.

Issue #2360, opened 2024-02-02
Gitter Sidecar dead / delaying page loading


abretaud:
Looks like Gitter sidecar is dead, https://sidecar.gitter.im/dist/sidecar.v1.js is no longer responding

Should it be removed?

On my firefox the hub pages stay for a veeeery long time in a loading state, waiting for [sidecar.gitter.im](https://sidecar.gitter.im/dist/sidecar.v1.js). And on usegalaxy.fr, for some reason the embedded home page stays blank until firefox gives up trying to load it.

dannon:
@abretaud Thanks for noticing this.  I'm surprised it's blocking and causing a delay since it should only be fired *after* the page loads, but I guess that doesn't matter because it's definitely dead now that Gitter is fully integrated into Matrix.  I'm removing the integration for now, and we can look for (or not) a new matrix-friendly option.

dannon:
There were actually three pages including sidecar directly as a part of the contents instead of injecting it as most of them did, which explains the blocking you saw.

abretaud:
Ok thanks!

PR #2363, opened 2024-02-05
Remove Gitter component

dannon:
Gitter has been fully integrated into matrix and the Sidecar integration we were using to display an inline chatroom is no longer available.  This was causing blocking loads on certain pages (see 176a89fdf16f096e0d46030ba9a751b54e8c2d51) that directly included the component, and would otherwise fail to load (but not block) everywhere else.

Down the road we will want to consider whether or not to shop around for or build and integrate a matrix-native alternative, but sidecar is a dead end and this removes it.

Fixes https://github.com/galaxyproject/galaxy-hub/issues/2360


bgruening (approved):
Thanks Dannon!

PR #2372, opened 2024-02-07
Add 23.2 release announcement

jdavcs:
Draft: need to replace ` TODOs` with images and videos. 

@natwhitaker how do we add images? I think the instructions in  `CONTRIBUTING.md` are outdated.


natwhitaker:
@jdavcs Hi! So I usually upload my images to the news file that I make (so for you, you could upload them to 2024-02-07-galaxy-release-23.2). And then to embed them into the index.MD you're writing, you would call them out by using content/news/2024-02-07-galaxy-release-23.2. Maybe that makes sense to you? I've always kept all the info needed for each news post contained in its own file - and I think that's what I've seen others do.

Issue #2374, opened 2024-02-07
SIG requirements - to who 'the nominated GCB representative' should be provided


lldelisle:
Would be good to explicitely give a contact to who 'the nominated GCB representative' should be provided:
https://github.com/galaxyproject/galaxy-hub/blob/22ecb045c0507c078e3a82187c6cc0c9f1f809f2/content/community/governance/gcb/index.md?plain=1#L62

bgruening:
Would it be bad to put this contact address on the homepage?

lldelisle:
No this seems good. I will try to clarify it.

PR #2375, opened 2024-02-07
invite to put the nominated GCB representative on the webpage

lldelisle:
See #2374 
@bgruening


bgruening (approved):
Yeah, this will work!

PR #2388, opened 2024-02-09
Update homepage 2 continuous integration fixes

hujambo-dunia:
@afgane These innocuous changes should fix the github webhook CI errors.


afgane:
For future reference, this is the source of the fix for Cypress failures: https://github.com/gridsome/gridsome/issues/249#issuecomment-625539828

afgane (approved):

PR #2390, opened 2024-02-12
Add caismd workshop

paulzierep:


bgruening:
I will merge, it might look like an empty page, in that case we need to add at least a short summary to it.

PR #2401, opened 2024-02-13
GEB Charter addition

nomadscientist:
Here is the exec charter added to the Hub site


nomadscientist:
Sorry folks, we're changing this PR to go here: https://github.com/galaxyproject/galaxy-hub/pull/2402

PR #2409, opened 2024-02-20
Fix for unsanitized URLs in the querystrings

hujambo-dunia:
Also fixes any potential phishing from undesirable Url redirects.


afgane (approved):
Thanks for addressing this so quickly!

PR #2411, opened 2024-02-21
Fix for misaligned US-EU labels on UseGalaxy button-dropdown redirect

hujambo-dunia:
Fix for misaligned US-EU labels on UseGalaxy button-dropdown redirect on Home page.

_Here is the issue we're hoping to resolve:_


src/pages/Index.vue
```diff
@@ -343,9 +348,12 @@ export default {
             if (priority[0]) {
                 galaxies.splice(0, 0, priority[0]);
             }
-            return galaxies;
+            this.prioritizedGalaxyLocales = galaxies;
```
dannon, line 351: Side effects in computeds need to be avoided

hexylena:
Is there a reason to go through /usegalaxy/redirects/, rather than going directly to the specified Galaxy? It seems like an unnecessary waste of 5 seconds for the end user. "Other" I guess could also go directly to /use/ which is a more informative page.

afgane:
The thinking was that if a user (especially a new user) simply clicked the button without taking note of the the US/EU/AU suffix, the intermediate page would bring it to their attention that there are multiple servers available and give them an opportunity to react. 

Your point about wasting 5 seconds of one's time stands so if the preference is to skip the redirect page, let's go with that option.

afgane:
On my dev setup, the US redirect is now taking me to `/US` vs. usegalaxy.org?

https://github.com/galaxyproject/galaxy-hub/assets/338642/2a5bc8e0-75ed-4d13-8e7f-f901bd22e4ec

hujambo-dunia:
@afgane Thanks for the QA! Yes, the first link (ie. "US") was still pointing to the Locale which I forgot to update (other links should've been working however).

> On my dev setup, the US redirect is now taking me to `/US` vs. usegalaxy.org?

afgane:
Same here, thanks. 

There's seemingly a relevant test failure still?

hujambo-dunia:
> Same here, thanks.
> 
> There's seemingly a relevant test failure still?

That `Cannot read properties of null (reading 'title')` error seems to be coming from the now deprecated (and deleted) `usegalaxy/Redirect.vue` file. Perhaps the CI just doesn't see the file is now deleted. I'll first just try re-running.

src/pages/Index.vue
dannon: Just gave this another look and it'd be cleaner to refactor this to be a computed without extra state set on mount, but it's not going to cause any immediate problems.

hujambo-dunia: @dannon thanks for the extra set of eyes on this! I previously had this as a computed on `master` in [this commit](https://github.com/galaxyproject/galaxy-hub/pull/2411/commits/54231fbc1060afd3299fe31979b2d7bf4a33738d#diff-75f76a7ed2a0f45f79ae96d58b3530c71c8b88cc3612a84ffd784ba8df2e3cbaL335-L346) but I was under the impression it was contributing to the labeling misalignment from the [gif video](https://github.com/galaxyproject/galaxy-hub/pull/2411#issue-2145760884) at the top of this PR so I removed it - my thinking was that using computed wasn't rehydrating the DOM with the updated compute values potentially because calling `getTimezoneOffset()` from one's browser was slower somehow on Production (the error wasn't fully reproducible on local). As long as the CI tests pass, I'm fine with merging without this change, but I'm still curious to know what you think the real culprit was for the misaligned-labeling issue? Perhaps my use of `[0]` referencing an Object from [this line](https://github.com/galaxyproject/galaxy-hub/commit/5ca4cecdb1d1fe46db6ec0759cc361d0a5e875de#diff-75f76a7ed2a0f45f79ae96d58b3530c71c8b88cc3612a84ffd784ba8df2e3cbaL18)..?

afgane (approved):
Barring the discussion on computed, lgtm.

Thanks.

src/pages/Index.vue
dannon: @hujambo-dunia I'll check out the busted ref and see if I notice anything, I'll go ahead and merge this as-is now, though, and can make a refactoring pass after.

PR #2417, opened 2024-02-23
Remove link to offline server.

ksuderman:
The NLP server at galaxy.lappsgrid.org has been offline for some time. We can add the link when (if) the server returns.


ksuderman:
I think all (most?) of their infrastructure is offline at the moment and because of the way the tools worked (calling backend SOAP and REST services) they were never published to a toolshed.  However, have your person contact me as the Lappsgrid recently gave Hopkins money to cover a few hours of my time.

PR #2418, opened 2024-02-26
update inputs for wf8

AnnaSyme:
Workflow 8 now requires input as `gfa`, not `fasta`. Also added detail about input being a single haplotype.


bgruening:
Thanks @AnnaSyme

PR #2464, opened 2024-03-15
Add Gigascience sponsor; add CFA to carousel

afgane:


martenson:
@afgane the owl is inviting us for GCC2025, that is next year (:

afgane:
Thanks for noticing that.

Updated in #2470.

PR #2473, opened 2024-03-19
Add a codeowners file, notify Natalie of news posts

shiltemann:
@natwhitaker this should make it so that any time somebody opens a PR that adds a news post, you get notified (either a GitHub notification and/or email notification, depending on your own notification settings)

Maybe it helps you with content for social media.


shiltemann:
oh wait, there already was a codeowners file, will move

bebatut:
Maybe also events?

Issue #2439, opened 2024-02-29
Fix for missing Title in Browser Nav Bar


hujambo-dunia:
_Currently says "Galaxy Community Hub - Galaxy Community Hub" which is duplicative; pre-redesign, one of those used to say "Home"._
![Greenshot 2024-02-29 14 30 18](https://github.com/galaxyproject/galaxy-hub/assets/3672779/2df7f0ad-2971-4435-bf59-d6d9a6d12149)

hujambo-dunia:
Fixed.

PR #2490, opened 2024-03-28
Fix for issue #2439

jai12789:
Changed the title of page from Galaxy Community Hub - Galaxy Community Hub to Home - Galaxy Community Hub


dannon:
@jai12789 Thanks for the interest in resolving this issue.  This isnt' quite the right approach here, as this would set the title statically for every page on the site.  This should be correctly resolved in #2562

PR #2493, opened 2024-04-02
Vazovn new leads

vazovn:
This is the cookbook pull request which we mentioned at the leader meeting on March 27th.


dadrasarmin:
To include the content of this PR on Galaxy-Hub we had to re-structure the directories. It has been done here. Thanks a lot for this contribution.

PR #2507, opened 2024-04-12
time added

erxleben:


bgruening:
@dannon do you know why the builds are failing? Is now in multiple PRs.

dannon:
@bgruening Thanks for the ping, I hadn't been watching these.  I'll take a look.

bgruening:
It seems that page load times have increased and cypress is complaining. I can confirm that loading the hub page on the usegalaxy.eu iframe feels also slower, which I thought is due to mother reasons... Maybe connected.

dannon:
@bgruening #2508 should fix this.  We can also test injecting that into the page head from the actual vue application, which would avoid this, but this is the quick fix.

PR #2508, opened 2024-04-14
Drop plausible (temporarily?)

dannon:
It's down and shouldn't be interfering with pageloads but this is, and is why the tests are failing currently.

todo: figure out why that's actually blocking when it shouldn't be, inject into head post-load maybe?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,7 @@
 
             gtag('config', 'G-KNS7RQJVRZ');
         </script>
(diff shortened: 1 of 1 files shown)


bgruening:
I see. I could not figure this out from the logs, but makes sense. I will enable it once plausible it back.

dannon:
@bgruening Sounds good to me.  When I have a minute I'll follow up on injecting that similar to other stuff we have done after the app actually loads, should be trivial.

Yeah, this would have been a pain to figure out exactly based on the github runner's logs, but running cypress locally (`yarn run cypress:open`)and looking at the console while it ran tests helped.

PR #2522, opened 2024-04-28
Add Vanessa's poster highlight from EGU2024

annefou:
Created a blog for the poster Vanessa Stoeckl presented at the EGU 2024 Conference.


bgruening:
Thanks a lot Anne. I moved it to the news section. The blog is dead IMHO.

PR #2523, opened 2024-04-30
Add blogpost with microgalaxy March hackathon outcomes and preprint

bebatut:


paulzierep:
Good from my side, this is basically the same as the preprint, right !

bgruening (approved):
Very cool!

PR #2528, opened 2024-05-03
Create index.md for EOSC symp

erxleben:


bgruening (approved):
I think the filename needs to have a date from today. Otherwise it's only displayed from that date onwards. Let's see if I'm correct.

PR #2555, opened 2024-05-20
News item for ARC deployment with IM

sebastian-luna-valero:
@maikenp please review


maikenp:
Thanks! This looks very nice for a new item. For the NeIC24 paper I think if we slightly modify the section " How to deploy ARC with IM?" and write in more of tell-how instead of a "tutorial" style, I think that would suit nicely.

bgruening (approved):

PR #2562, opened 2024-05-24
Fix for SEO to properly list 'Home' in homepage metadata

hujambo-dunia:
Addresses #2439


src/pages/Index.vue
```diff
@@ -363,11 +363,20 @@ export default {
     mounted() {
         this.setPrioritizedGalaxyLocales();
     },
+    metaInfo() {
+        return {
+            title: this.$page.main.title,
```
dannon, line 368: It's much cleaner and more straightforward to just return `Home` here directly, skip the .md and page query.  There's just not a good reason for the obfuscation.

PR #2566, opened 2024-05-24
Fix for US-EU daylight savings

hujambo-dunia:


dannon (approved):
Probably a library we can include to automatically do this, so we aren't doing it every 6 months?  Future problem, though!  :)

PR #2568, opened 2024-05-28
News item on Check-in avaliable in UseGalaxy.eu

enolfc:
A short news item on the availability of Check-in as login option for UseGalaxy.eu


bgruening (commented):
Just checked the DB and currently 55 users have used EGI check-in.

@enolfc could you maybe add also the "internal option". If you already have a Galaxy account, but would like to connect Galaxy account to EGI, you can go to user-preference and click the button .... etc... maybe with screenshot as well?

Thanks a lot!

enolfc:
It seems rendering is not good, submitting a new PR to fix

PR #2576, opened 2024-06-03
Add gtn integration highlight article

ElectronicBlueberry:


bebatut:
What about having this news in the GTN first and then imported here after?

hexylena:
> but I think as a new feature to Galaxy that users will see in the interface it's more Galaxy News than GTN News?

Yep. But then we'll have to import it back into the GTN because it's of interest to our audience as well, and it'll be potentially more visible for folks who are able to use when it is in the GTN.

Agreed it's a difficult balance, I guess with #2578 we're trying to simplify it to "is this GTN relevant? Post to the GTN. done." and then all of our extensive automation takes care of the rest (distribution to the hub, matrix:gtn/goats/hub_social, several slack locations, mastodon.)

bgruening:
For what it's worth there will be a second post about Galaxy - helpforum integration, both go hand in hand.

I will stay out of this discussion now, it's not beneficial for my mood. Please just make sure we get this content published.

dannon:
Yep, I think there's a longer conversation to be had here, but let's just move it forward for now.  Everyone's very busy for at least the next 4 weeks!

PR #2578, opened 2024-06-04
Swap to simpler implementation of GTN news import to hub

hexylena:
@dannon I'm guessing you probably had something similar but we need this because asking people to cross post is too difficult, so I rewrote it myself. At least then GTN news can go in the GTN and be automatically imported (since all gtn news is hub news but not all hub news is gtn news)

- removes complicated API logic
- just clone, run script, commit, done


bgruening:
I guess the other way around would also be useful, import to GTN from the Hub.

hexylena:
@bgruening the other way around is more complicated and would not produce as nice results, though:

- All GTN news is hub news but not all hub news is GTN news.
- The GTN has _strictly_ better metadata (usernames mapped to human names/other json-ld metadata, not a free text field)
- The GTN has an RSS feed that's easy to parse, the hub lacks an RSS feed. We'd be stuck cloning the entire hub to pull out posts.

hexylena:
In the long term, if the hub gets an RSS feed, I would love to be able to import news articles, absolutely. especially if they have sufficient metadata for the GTN. :)

hexylena:
(@shiltemann reminds me to copy and paste this script for events.)

dannon:
@hexylena Yeah, similar, but no worries -- I didn't manage to wrap it up before being out on holidays.  But I'm back now, so happy to merge this!

I don't mind this being one directional -- the hub is intended to be the aggregator and I don't know that having the GTN also aggregate helps anything.  That said, there are existing json feeds for news, events, and platforms -- even easier than RSS.

hexylena:
Oh, didn't know there was a json feed, that's great to know :)

shiltemann:
> it will be pretty annoying to tell people after they created a news item here to do it again in another repo.

the idea is that people would add GTN related news to the GTN, and then get the Hub for free, so they still only have to do it once (sure, we may have to do some communication there if we find a lot of people adding GTN new to Hub first but)

PR #2653, opened 2024-06-20
Update index.md

subinamehta:


subinamehta:
The name was wrong, so just corrected it

PR #2659, opened 2024-06-23
Switch workflow talks

mvdbeek:
@afgane is that all that's needed ?


davelopez:
I wonder if it also makes sense to change the demo session from Monday to Tuesday if there is a free spot on Tuesday

afgane:
We updated the presentation schedule slides as well so should be sorted. Can I ask why this change request so late in the process?

We can sort the demo over lunch. The on-site schedule has been printed already so in case folks reference that one, we'll need to make a note it's being moved as well.

PR #2691, opened 2024-07-10
add call for trainers, remove carousel

mira-miracoli:
image


content/bare/eu/usegalaxy/main/index.md
```diff
@@ -4,8 +4,11 @@ title: Galaxy Europe
 
 <slot name="/bare/eu/usegalaxy/notices" />
 
-<Carousel />
```
bgruening, line 7: Can you leave this in, but comment it out?

PR #2722, opened 2024-08-03
EuroScienceGateway news item about WorkflowHub Publisher forum

stain:
... not sure which `subsites` to list as it's not Galaxy-specific.


bgruening (approved):
This looks good, also subsite should work. Thanks @stain

PR #2723, opened 2024-08-03
Update Makefile for latest conda and yarn packages

stain:
Note that you may have to do `make build` twice for some reason. Afterwhich `make run` works fine, at least on Mint 21.3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ create-env: ## create galaxyproject-hub conda environment
 	if ${CONDA} env list | grep '^${CONDA_ENV}'; then \
 	    ${CONDA} env update -f environment.yml; \
 	else \
-	    ${CONDA} env create --force -f environment.yml; \
+	    ${CONDA} env create -f environment.yml; \
 	fi
 .PHONY: create-env
 


bgruening:
I will that one for @dannon not sure about the lock file update.

dannon:
I need to dig up why we had to pin sharp back to `0.30.0` -- hopefully I left notes...

martenson:
the removal of `--force` is a needed change, this fails on newer conda versions I think
```
conda env create --force --file=environment.yml
/opt/homebrew/Caskroom/miniforge/base/lib/python3.12/argparse.py:2006: FutureWarning: `remote_definition` is deprecated and will be removed in 25.9. Use `conda env create --file=URL` instead.
  action(self, namespace, argument_values, option_string)
usage: conda [-h] [-v] [--no-plugins] [-V] COMMAND ...
conda: error: unrecognized arguments: --force
```

scottcain:
> I need to dig up why we had to pin sharp back to `0.30.0` -- hopefully I left notes...

I know sharp caused me problems on my M1 Mac, but I doubt that was the issue for you. I did successfully install sharp on my M4 Mac, but it took a little hoop-jumping-through.

bgruening:
@dannon can we merge this or can I create a very similar PR. It builds locally and everyone that builds it locally will have those changes now.

dannon:
@bgruening I'll verify it still works locally in other envs and will resolve the conflicts, and we can merge, sure.

dannon:
Trying to debug, but this actually does break the build for me.  I'm going to split off a separate PR with just the --force / -f bit.

dannon:
Sorry this took forever -- I was hoping to resolve the dependency issues here but I think that's a larger issue and I don't want to hold up the conda fix.  I'll revisit the dependency issues (lock file changes) separately.

PR #2724, opened 2024-08-06
Update Galaxy Australia platform description

PatCapon39:
Extensive updates to this page: https://galaxyproject.org/use/usegalaxy-au/

- new screenshot
- new comments section
- expanded support options
- clarified disk quotas
- updated citations section
- updated sponsors


PatCapon39:
@GarethPrice-Aus can you please see if you agree with the wording of the Gitter support channel? The rest should be non-controversial
Ps this won't fix the weird formatting that is going on for all of the usegalaxy.* platform descriptions, I don't know why that's broken.

PR #2727, opened 2024-08-07
Fix style leakage -- for simplicity, not replacing it with localized …

dannon:
…styles since this event is passed and folks aren't likely glued to it on their phones anymore.


bgruening (approved):
Thanks Dannon!

Issue #2721, opened 2024-07-31
Image links are wrong in the feed


arash77:
It seems that the Atom feed generated at https://galaxyproject.org/eu/feed.atom contains incorrect image URLs, causing the images to not load properly.
For example, in the post ["What do you want to know about Galaxy-EU? Our first 'Ask us anything' Q&A session"](https://galaxyproject.org/news/2024-07-29-ask-us-anything-report/), the image URL is currently:
`https://galaxyproject.org/news/2024-07-29-ask-us-anything-report/questions_1.png`
However, it should be one of the following:
* `https://galaxyproject.org/assets/static/questions_1.175c05c.0caec253b2ef2e0beaac0cb1f447d02c.png`
* `https://raw.githubusercontent.com/galaxyproject/galaxy-hub/master/content/news/2024-07-29-Ask-Us-Anything-report/questions_1.png`

This issue was first identified while using the Atom feed to create automated Galaxy social PRs: https://github.com/usegalaxy-eu/galaxy-social/pull/86#discussion_r1697182311.

Could you please look into this?

Thank you!

bgruening:
@dannon do you have a clever idea here?

arash77:
I realized that images not relatively addressed (external images) in the Markdown, remain unchanged. For example, in this post:
https://galaxyproject.org/news/2024-07-08-galaxy-social/
The image mentioned in the Markdown is:
https://github.com/arash77/galaxy-hub/assets/2973722/c8de8bdf-3cee-466f-aea3-d2462b4bc88f
This image remains consistent across the website and in the feed.

dannon:
Alright, I finally had a minute to look at this, sorry for the delay @arash77, @bgruening

The staging I was referring to (out of band in either dev meeting or chat, wherever) is here:  https://github.com/galaxyproject/galaxy-hub/blob/master/src/build/run.mjs#L95.  This is working.  It stages the source file `content/news/2024-07-29-Ask-Us-Anything-report/questions_1.png` to `dist/content/news/2024-07-29-Ask-Us-Anything-report/questions_1.png` at the end of the build, for direct access.  That dist dir is what we end up publishing as galaxyproject.org.

Note the case.

https://galaxyproject.org/news/2024-07-29-ask-us-anything-report/questions_1.png

Which is generated in the feed, fails, but the following URL exists and works.

https://galaxyproject.org/news/2024-07-29-Ask-Us-Anything-report/questions_1.png

[...]

PR #2735, opened 2024-08-14
Adjustments to static file staging

dannon:
xref #2721

Checking to see if this passes the smoke test, no (newly) broken links, etc.


dannon:
Ahh, looks like we rely on the domain and path being lowered, but never do so with the actual filename -- adjusting now.  Glad we added those broken link checkers!

bgruening:
Cool, thanks a lot Dannon

PR #2741, opened 2024-08-19
Adjust static file staging location handling

dannon:
Uses lodash.kebabcase to handle paths like:

```
news-2024-08-14-GalaxyInResearch/GalaxyInResearch_Jain2024.png
Which should be:
news-2024-08-14-galaxy-in-research/GalaxyInResearch_Jain2024.png
And not the simple lower(), like:
news-2024-08-14-galaxyinresearch/GalaxyInResearch_Jain2024.png
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "ics": "^2.35.0",
     "jiti": "^1.12.0",
     "lodash": "^4.17.21",
-    "lodash.kebabcase": "^4.1.1",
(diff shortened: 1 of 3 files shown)


dannon:
Frustratingly, this still doesn't match the standard.  We kebab-case CapitalizedNames, but not between numbers as in `gcc2024`.  Another iteration.

PR #2758, opened 2024-09-06
Adjust static staging path generation (again)

dannon:
Should address https://github.com/galaxyproject/galaxy-hub/issues/2721#issuecomment-2333778796

Let's make sure tests run to look for broken links.  And on that note, I'll follow up separately with testing the feed as well as the navigable site for broken links, so we catch all these.


bgruening:
Thanks @dannon!

PR #2780, opened 2024-10-01
Galaxy Mass Spectrometry Community - SIG Page first version

hechth:
This adds an initial version of the updated Galaxy Mass Spectrometry Community SIG page.

I am aware that the calendar and mailing list are currently still missing but it is something that we will add probably in the next year. Also so far, the old SIGs etc. are still in place and their mailing lists still work.


bgruening:
Thanks @hechth!

PR #2789, opened 2024-10-10
adress the number, see mail from World Animal Foundation

bgruening:


scottcain:
@bgruening maybe "74,962 and counting" because is the number changing on a regular basis?

scottcain:
Oh, I see, this is a count of *identified* vertebrate species (like, in the whole world), which I suppose doesn't change that often. I thought this was a reference to a number of assemblies.

bgruening:
Thanks @scottcain!

PR #2813, opened 2024-11-15
create new post - MGnify

RZ9082:
Creating a news post for the MGnify amplicon v5.0

@paulzierep @bgruening


bgruening:
@RZ9082 if you like you can embed a workflow preview into the post and you could also add a `run workflow` button if you like.

PR #2825, opened 2024-11-19
updating recent publications

scottcain:
Note that there was some reformatting introduced by neovim. If that's annoying, I can probably undo it.

### DIFF
--- a/content/home/research.yml
+++ b/content/home/research.yml
@@ -1,49 +1,49 @@
 heading: Galactic pulse
 
 research1:
-    title: Recent publications
-    content: "Galaxy can be installed locally, fine-tuned with a tailored toolset and no usage quotas."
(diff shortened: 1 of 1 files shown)


bgruening:
Can you have a look at the failing tests? We should stay consistent about our YAML styles, which one we take I don't care much.

scottcain:
@bgruening is that what the problem is (yaml style)? It was unclear from the error messages I was seeing. I'll try repairing it and committing again.

dannon:
@scottcain The test failure is unrelated to the yaml formatting, but it's something we can fix.  I'm opening a PR that'll just autoformat yaml a single consistent way (with `yarn run format`) across the codebase and then we don't ahve to worry about any inconsistency of local editors.

PR #2839, opened 2024-11-27
Update Publish workflow for improved efficiency

arash77:
Upgrade `publish.yml` to the latest action versions and simplify the caching process in the workflow. Based on [setup-node docs](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data).
Also, this is done as `cache@v1` is [deprecating](https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/).


bgruening (approved):
Thanks!

PR #2843, opened 2024-11-27
Add news item about TACC/CFDE grant

scottcain:


jdavcs:
For consistency's sake, should we change the directory name to include the date? i.e., `2024-11-tacc-sdsc-grant` -> `2024-11-26-tacc-sdsc-grant`? I'm sure it'll display correctly in any case.

PR #2852, opened 2024-12-03
Set up basic gbcc2025 redirect

dannon:
@afgane We can also add a toplevel galaxyproject.org nginx alias if you want once the site is live for an immediate redirect instead of the javascript one if we want it a little cleaner.


afgane:
I'd think this is fine. Hopefully most links and folks will go direct to gbcc2025.org so this is primarily for the future archive.

PR #2890, opened 2025-01-15
Update index.md of BFSP-FairyMAGs news

dadrasarmin:
The images were not loaded. I think now it should be visible.


dadrasarmin:
I am not proficient in HTML tags and syntax, but the pattern I see is as follows:
This does NOT work:
```
<figure>
<img src="file.png" alt="second_name"/> 
<figcaption>Some text</figcaption> 
</figure>
```
But this works:
```
![second_name](file.png)

</div>  
<figcaption>
  Some text
</figcaption>
</div>  
```
All the links that were in the figcaption, besides this one, would show the image.

bgruening:
ok, lets use:

```
<figure class="figure">
  <g-image src="./file.png" class="figure-img img-fluid rounded" />
  <figcaption class="figure-caption">Some text.</figcaption>
</figure>
```
Thanks for figuring this out.

PR #2910, opened 2025-02-04
update all pages that mentioned old status page

martenson:
- the old one does not update anymore and there is no path for fixing that atm
- I refrained from changing news items with the old link


martenson:
alternatively we could just redirect status.galaxyproject.org there, thoughts @natefoo?

bgruening:
I prefer the redirect. The new URL no one can remember :)
A gxy.io link would also be good imho.

natefoo:
Does UptimeRobot offer custom domain functionality? I can redirect but it'd be much nicer if it could actually just be status.galaxyproject.org as before.

Also I guess I missed the discussion on statuspage not being fixable.

martenson:
@natefoo I ranted about it a couple of times but seems not enough to reach you. 🤣 

tldr: we've always used uptimerobot for doing the actual monitoring, it used to send emails to atlassian status page to update status. 'Bout year ago uptimerobot changed business model and disabled free monitoring, I applied for open source free plan and they recently gave it to us, so we can monitor nicely and we can have status page on their site. However they do not support custom domains in our plan and do not support sending emails for status changes.

So we can e.g. 
- do with what status page they give us
- pay them to get custom domain
- write our own integration of sorts to use their status tracking for updates of atlassian page
- abandon uptimerobot and find someone else (I didn't, but there could be someone better)
- write our own damn status tracking app (with blackjack)

martenson:
afaik statuspage is still just a "static" website with no monitoring https://support.atlassian.com/statuspage/docs/know-when-to-automate-your-status-page/

natefoo:
Ok I set it up as an iframe running on one of our EC2 instances, what do you think? https://status.galaxyproject.org/

dannon:
Looks great to me!

martenson:
@natefoo lgtm, let's go with that!

PR #2951, opened 2025-02-20
modify EU storage

pavanvidem:
The article reads well and super clear. But I have a few concerns on the terminology used in unshared storage which led to confusion between storage and quota.

- The unshared storage is not really a separate storage, it is part of regular long-term storage. On the contrary, the short-term, user-owned are real storages. On the storage dashboard, does user see unshared storage as a separate entity?
- Is this statement true? "Data on this storage cannot be used for public datasets"
- Should we add storage of sensitive data as a usecase for unshared storage?

As I see, there 3 quota classes, long-term, short-term and user-owned. There are sharable and unsharable storage (or space) of long-term user quota.

### DIFF
--- a/content/eu/storage/index.md
+++ b/content/eu/storage/index.md
@@ -204,7 +204,7 @@ The data is stored on a high-available data storage and you are allowed to share
 ## Short term storage
 
 S3-based object storage is maintained by the compute center of the University of Freiburg.
-This storage, also called scratch-storage, is `purged monthly` and so it is only appropriate for short-term methods development and such.
+This storage, also called scratch-storage, with data `purged` after 60 days after creation and so it is only appropriate for short-term methods development and such.
 The rapid deletion of stored data enables us to provide this storage without a quota. This storage is not backed up.
 
 The automatic cleaning of this storage works like this:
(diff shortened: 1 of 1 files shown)


bgruening (approved):

bgruening:
> The article reads well and super clear. But I have a few concerns on the terminology used in unshared storage which led to confusion between storage and quota.
> 
>     * The unshared storage is not really a separate storage, it is part of regular long-term storage. On the contrary, the short-term, user-owned are real storages. On the storage dashboard, does user see unshared storage as a separate entity?

Technically it will be a separate storage. An additional S3 bucket. But the quota spans two storages.

>     * Is this statement true? "Data on this storage cannot be used for public datasets"

Yes, I think so.

>     * Should we add storage of sensitive data as a usecase for unshared storage?

It is meant for this use-case - maybe in the future. But we should not advertise it as long as our terms do not allow sensitive data.

> As I see, there 3 quota classes, long-term, short-term and user-owned. There are sharable and unsharable storage (or space) of long-term user quota.

Its really 4 options. `unshareable` is a separate bucket. We probably want to safeguard those private data in a special way and not mix it with the other data in long-term storage.

pavanvidem:
> Technically it will be a separate storage. An additional S3 bucket. But the quota spans two storages.
> 
> >

Then I misunderstood it. All good.

PR #2966, opened 2025-02-26
Better date support for importing events from GTN RSS feed

arash77:
Using [the updated RSS feed](https://github.com/galaxyproject/training-material/pull/5799) for events to receive date, duration and external tag.
Also figured out there isn't any tag related to the events in GTN events RSS feed so that has been removed.


bgruening:
Thanks a lot!

PR #2986, opened 2025-03-11
Add blog post with Invenio/Zenodo integration update

davelopez:
Some updates about the new way of integrating user file sources in the context of Invenio and Zenodo.

Also updates the old post to point to the new one.


bgruening (approved):
Thanks a lot!

Issue #2991, opened 2025-03-13
Add cross-links to annual conference on internal pages


hujambo-dunia:
1. https://galaxyproject.org/gcc/  "update with the report from 2024 and a pointer for 2025"
2. https://galaxyproject.org/news/2024-09-03-gbc-c2025/

PR #2993, opened 2025-03-13
Added recent and upcoming GCC conferences

hujambo-dunia:
Addresses #2991


hujambo-dunia:
@nakucher Let me know if the [value in the "Hosts" column here](https://github.com/galaxyproject/galaxy-hub/pull/2993/files#diff-c2f3c81cb4bb1e8efcb46716bef4ee719c1aeb65ff127ac72b82b57e9da38fb5R40) which I list as `CSHL` is accurate.

nakucher (approved):
Looks great @hujambo-dunia!

PR #2994, opened 2025-03-13
Add gbcc gta homepage

scottcain:
Thought I made a PR for this hours ago :-/


scottcain:
@dannon checks are failing because of prettier, but I ran prettier with the defaults and it still failed. Makes me think perhaps there is a .prettierrc somewhere that I might want to have. Any idea?

scottcain:
OK. running `npm run format` resulted in a fix, but it is a mystery to me why that is different from running `prettier --check` on the offending file.

dannon:
Might be `prettier` check wasn't automatically including `.vue` files without the explicit inclusion in the package script?  System version?

Either way, this looks good to me, thanks!

What do you think about dropping the green box at the top?  It seems out of place to me and the content overlaps (not on screen, but in content) with the new images added here.

dannon (approved):

PR #2998, opened 2025-03-14
Sig page - update index 

nomadscientist:
Screenshot 2025-03-14 at 12 55 47

As per our previous GCB meetings, I have simplified the SIG index page and updated the cards to make it clearer they are links. Then SiGs can put all their information (and keep it updated) on their required SiG page, rather than on both the index and their SiG page.


nomadscientist:
I have 0 idea on that test error, help would be appreciated!

PR #3013, opened 2025-03-20
Add usegalaxy.fr to UseGalaxy button

bebatut:


hujambo-dunia:
@bebatut Made a "short term recommendation to your PR. Long term, an IP geolocation service/package (without the intrusive "share your location?" pop-up windows) would be the best long-term option for everyone. There is an open source IP geolocation package called [GeoLite by MaxMind](https://dev.maxmind.com/geoip/geolite2/) which some of our existing third-party services use (Sentry, Plausible -but Plausible doesn't expose it to the frontend). There might be other use cases where we need country exposed to the front-end which @dannon has already researched even so I'll bring it up at the next ui/ux meeting.

PR #3061, opened 2025-04-04
Use 'authors' instead of 'author'

RZ9082:
Just noticed multiple blog posts using 'author' instead of 'authors'.


bgruening (approved):
Thanks.

PR #3076, opened 2025-04-09
DO NOT MERGE trying out adding a new graphql generated page

scottcain:


scottcain:
@dannon would you mind taking a look at this and telling me what else is needed? Here's what I'm trying to do: I was given an example of a page that might be nice a table of news items that have "paper" as a tag, so I created a graphql query that (I think) returns the appropriate result, and then I tried adding it to the build, but I don't think it works (going to /newspapers when I'm running locally results in 404) so I probably need some more config that isn't as obvious to me.

Note also that I don't really want to add this to the production site, I just want to learn how to do it. Thanks!

dannon:
Sure, I will take a look. Not sure I can get to it today but I will try

dannon:
Sure, I will take a look. Not sure I can get to it today but I will try.  Glad to see you experimenting with this!

dannon:
@scottcain sorry this took forever for me to pull and run.  If you want it to be at /newspapers/, move the file to Index.vue in a /pages/newspapers/.  Right now for non-index named components we render a page based on the name, so this would be available at http://localhost:8080/everywhere/news-papers/ locally. (and it does work!)

bgruening:
@scottcain we can close this experiement, do we?

PR #3078, opened 2025-04-10
Show all events on the ESG page

bgruening:
Unfortunately, the filter is not showing all events and there is also no way/link to get all events.

In addition, ESG is soon over, so I assume no new events/posts will be added, in this case the events page will be empty in a years time.

I have removed the filter, I hope that is ok.

### DIFF
--- a/src/pages/projects/esg/Events.vue
+++ b/src/pages/projects/esg/Events.vue
@@ -86,8 +86,7 @@ query {
 
     recent: allParentArticle(
         sortBy: "date", order: DESC, filter: {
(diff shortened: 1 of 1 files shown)


ElectronicBlueberry (approved):

dannon:
Seems fine but which ones are not returned in the filtered query?  We should see what's going on.

bgruening:
This PR was just fixing the "how much is displayed" on the site. I guess the filters are a different thing? I will comment on your PR.

PR #3079, opened 2025-04-10
Updating the event section of the EuroScienceGateway project

dadrasarmin:
I noticed two problems at EuroScienceGateway subsite.
1. When I navigated to the events page, there were many events that are not related to ESG directly, however, they are related to the Europe subsite.
2. I can only look at the events for the last 12 months.

I changed the a few things to address both. It seems fine when I build it locally.

Maybe for future: I feel there are meetings and events related to EuroScienceGateway that are available on Europe subsite and we have to find them and add metadata to them.

### DIFF
--- a/src/pages/projects/esg/Events.vue
+++ b/src/pages/projects/esg/Events.vue
@@ -86,8 +86,7 @@ query {
 
     recent: allParentArticle(
         sortBy: "date", order: DESC, filter: {
-            category: {eq: "events"}, subsites: {contains: ["esg"]}, draft: {ne: true},
-            has_date: {eq: true}, date: {between: ["2022-07-01", "2025-12-30"]}
+            category: {eq: "events"}, subsites: {contains: ["esg"]}, draft: {ne: true}
         }
     ) {
         edges {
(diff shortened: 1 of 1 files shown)


src/pages/projects/esg/Events.vue
```diff
@@ -86,7 +86,8 @@ query {
 
     recent: allParentArticle(
         sortBy: "date", order: DESC, filter: {
-            category: {eq: "events"}, subsites: {contains: ["eu"]}, draft: {ne: true}
+            category: {eq: "events"}, subsites: {contains: ["esg"]}, draft: {ne: true},
+            has_date: {eq: true}, date: {between: ["2022-07-01", "2025-12-30"]}
```
bgruening, line 90: I'm not sure why we should filter by date. We want to display everything with tag ESG isn't it?
A date based filter only makes sense when we also have some `pagination` in place to also scroll to the other out-filtered posts.

bgruening, line 90:
```suggestion
            category: {eq: "events"}, subsites: {contains: ["esg"]}, draft: {ne: true}
```

dadrasarmin: Ok. I see your point on this. For me, it was educational to pick specific dates. :)

PR #3089, opened 2025-04-17
removing the reg link for GCC2024 and adding GBCC2025

scottcain:
@bgruening I noticed that https://galaxyproject.org/eu/ had a link to GCC2024 registration, so I removed it and replaced it with a reg link for GBCC2025.


bgruening:
Mh, how did this happen. Thanks for the catch!

PR #3096, opened 2025-04-25
added logos to the esg Workflow event

teresa-m:
added logos to the esg Workflow event


bgruening:
Lets see, I will fix it when it does not look correct.

teresa-m:
Maybe the logos are a bit too close. But I think it is ok: https://galaxyproject.org/events/2025-05-13-byow-workshop/

PR #3104, opened 2025-04-28
Create event page for Galaxy Workshop on Digital Humanities in Freiburg

Sch-Da:
Creating an event page for a Galaxy Workshop in Freiburg


Sch-Da:
thanks for improving this!

Issue #3124, opened 2025-05-08
Updates for GTN event import bot


shiltemann:
@arash77 thanks again for making the importer! Today I noticed some suboptimal behaviour:

There was a GTN external event, that linked to an event page here on the Hub. The importer then tried to import this event, and overwrote the existing event (ending with a circular reference of GTN pointing to Hub, pointing back to GTN)

Suggested solution:
- if the GTN event has `layout: event-external`
- check if the `url: ` parameter is a Hub link, and skip the event if so

Secondly, for other `event-external` GTN events, it would be best if the Hub linked directly to that URL, rather than to the GTN.

arash77:
Thanks for the feedback! External URLs aren’t available in the feed, but we can use the `already-on-hub` tag to skip importing those events for now.
I will try to find a way to include the direct external URLs.

PR #3127, opened 2025-05-09
Fix slug generation in gtn import

arash77:
Ensure that slug generation correctly handles URLs by removing any trailing slashes before processing. This is particularly important for external links, which may not follow the same structure as GTN links. Additionally, skip external events that link to Galaxy Hub events to avoid duplication.
This addresses https://github.com/galaxyproject/galaxy-hub/issues/3124.


bgruening (approved):

bgruening:
Thanks @arash77

PR #3128, opened 2025-05-12
EGD2025 update schedule

dadrasarmin:
Hi,

I have added a draft schedule to the event page and removed the accommodation section. Oana received a few offers which were more expensive compared to the normal bookings and we decided to not include them.

(@bgruening and @hrhotz) Could you please look at the draft schedule? I am open to change it. However, I think it is good to shape it based on the registration information that we receive slowly. For example, I noticed that the current schedule has significantly more male than female or diverse volunteers to present something. I would like to change it and include more female/diverse scientists.

Moreover, I think we need more registration from ELIXIR communities. That section is almost empty.


hrhotz:
Hi Armin

Thank you very much for moving forward. It is always good to have a preliminary schedule up as soon possible, as this will attract more registrations...hopefully.
I do see your concern about the 'dominance' of the old white men in the schedule. We were in similar situations at previous occasions in the build up for the event. However, again looking back at previous occasions, it actually leveled out pretty well in the end (most of the time).
I would not be too much concerned at this moment, let's wait (at least till after GCC) till we have more registrations.
If it looks really bad, we can always approach people and try to convince them to present something, and thereby shift the balance.

Otherwise, I had a very quick look at your suggestions and they look perfect to me

hrhotz (approved):
see my comments above

dadrasarmin:
Thank you very much for your feedback. You are correct GCC could help us to get more registration and reach more people.

PR #3130, opened 2025-05-12
Port article tag filtering to main news page

ElectronicBlueberry:
Ports the tag filter feature from the esg subsite to the main news page, allowing dynamic links like: `https://galaxyproject.org/news/?tag=TIaaS`


bgruening:
Thanks @ElectronicBlueberry!

PR #3138, opened 2025-05-14
Revert "Import GTN Events"

shiltemann:
Reverts galaxyproject/galaxy-hub#3125


shiltemann:
@bgruening same problem as before, circular linking to the Digital humanities workshop between GTN and Hub with the full information being in neither place.

bgruening:
arg! Thanks

PR #3143, opened 2025-05-15
add anylabeling news

sunyi000:
news about AnyLabeling IT in Galaxy Europe
@bgruening @beatrizserrano


content/images/galaxy-anylabeling-results.png
bgruening, line 1: is that a goejson? Can you visualise that in the geojson viewer in Galaxy?

sunyi000: Clicking the Visualization button gave me options of Cytoscape, Editor, Graph Visualization and Vitessce, only Editor can be used.
I also there's a CSG viewer, but it cannot be used either.
Yi

bgruening: Turn your filetype into `geojson` not `json` (if it is a geojson) than it should work.

bgruening: Oh too bad :) That would be nice win!

dadrasarmin (commented):
Thanks this looks very good. It would be nice to enrich the metadata of pages for the Galaxy Hub. We are planning to use them more in future.

Issue #2524, opened 2024-04-30
broken link/feature after EU's migration to the hub


hexylena:
Noticed the following URL cited in the TIaaS paper, clicking on it redirects to the general news page rather than articles tagged TIaaS.

https://galaxyproject.eu/news?tag=TIaaS.

thankfully exists in way back machine for a potential reader, but, it would be nice not to break URLs.

Issue #3001, opened 2025-03-14
Create a template for writing news/events on the Galaxy hub


dadrasarmin:
## Background

At the moment, the following resources are available to write an event, news, and blog post on the Galaxyproject webstie:

- [About The Galaxy Community Hub](https://galaxyproject.org/hub/)
    - [Contributing content](https://galaxyproject.org/hub/contributing/)
    - [Advanced Markdown](https://galaxyproject.org/hub/contributing/markdown/)
    - [The Global Hub](https://galaxyproject.org/hub/global/)
- [Galaxy Community Hub](https://github.com/galaxyproject/galaxy-hub/blob/master/CONTRIBUTING.md) which is not on the https://galaxyproject.org/ 
    - Specifically the [Page metadata](https://github.com/galaxyproject/galaxy-hub/blob/master/CONTRIBUTING.md#page-metadata) is related to what I am going to describe below.
        - General metadata: title, autotoc
        - Section specific metadata
            - News item metadata: title, tease, date
            - Event item metadata: title, date, days, tease, continent, location: name, location: url, external_url, image, gtn, contact, tags, subsites
            - Blog item metadata: title, tease, date, authors, external_url, source_blog, source_blog_url

## Problem

I think for some of these metadata options, the valid options are not clear for new contributors. For example, I was not sure what are the valid sub-site keywords, which I found later [here](https://github.com/galaxyproject/galaxy-hub/blob/master/config.json). Also, `supporters` tags will not work with the current website but it is widely used in many events and news posts ([open issue here](https://github.com/galaxyproject/galaxy-hub/issues/1612) since 2022). Finally, I believe that we do not have a collection of tags standardized that we are all encouraged to use. Therefore, each contributor will use what they think is the best on the go, however, if it becomes standardized or there will be a suggestion system the FAIRness of the contents will improve. Therefore, these `tags` then could be used by the users to retrieve information related to a specific keyword.

dannon:
I thought we had a few examples showing available fields, but if they're not findable then we should absolutely do better!  Pre-filled templates for the common types that can be copied in and edited would be fantastic.

dadrasarmin:
@arash77 suggested to create templates for news and events as mentioned above. It would be nice for galaxy-social to include photos (with a default option). I have prepared two drafts:
1. News template:
```
---
title: "Your Title"
tease: "Short bit of additional text about the news item."
authors: "Your Name" 
date: "YYYY-MM-DD"
[...]

Issue #2864, opened 2024-12-18
Dream SIG Page widgets


nomadscientist:
In my magical dream world, each SIG page has an automated chart with the following data:

- number of members, events, publications, tools, training, news posts, per year as interactive charts, all hosted on the SIG page

Obviously this means we need to tag stuff effectively, but if we (as SIGs) make sure that we do, the payoff of such automated metric tracking would be excellent.

This means:
1 - Obvious place to copy links to papers + a tag/SIG tag (i.e. spoc) for them
2 - Tools I suspect could be linked to toolshed topics? (for us, single-cell & spatial omics, so two tags, but microbiome will be more complex, for example)
3 - Training, news, & training events are all tagged/collected by GTN already so that data could be used
4 - News/events not in the GTN can be tagged and pulled?

Members is the tricky one - but at a minimum, the GTN contributors page has options for metadata on editorial boards, so could include membership potentially as well?

scottcain:
Your comment about tagging "non-GTN" content (which I assume is to say, "stuff on galaxyproject.org"). Having some way to do that would be awesome, but it's not clear to me if that tech is readily available with the current gridsome/vue system (is it? I really don't know).

If it's not currently possible with gridsome/vue, we should consider migrating to something more fully featured.

nomadscientist:
That would be ideal - it would be awesome if the GTN content, tool info, and community activities (Hub) could be integrated per domain, to make truly cohesive community subdomains/pages that all community members can contribute to and find useful. This likely involves GTN + CoDex + Hub + IWC + more

Issue #2200, opened 2023-10-05
Events horizon searching feature


nomadscientist:
I want to be able to search #singlecell and see any events tagged with #singlecell
Then I can embed this search into the single cell training & community pages, and thus keep users updated on relevant events.
AUTOMATICALLY

Can this feature exist please?

Shashankpantiitbhilai:
assign this to me

PR #3161, opened 2025-05-22
converted TIaaS to lowercase for tag search.

dadrasarmin:
There was a [PR](https://github.com/galaxyproject/galaxy-hub/pull/3130) by Laila to add search tag feature. Unfortunately, it did not work with https://galaxyproject.org/news/?tag=TIaaS but worked for other searches such as https://galaxyproject.org/news/?tag=paper . Our guess was the case sensitive search. For this specific case, I changed the files that had TIaaS as a tag with the following search: `grep -FRl 'TIaaS' content/events/ --include='*.md'`. I have added it for a few posts that the tag was missing. It works now when I built the website locally.

I believe this PR will fix [this issue](https://github.com/galaxyproject/galaxy-hub/issues/2524).

### DIFF
--- a/content/news/2020-07-30-galaxy-africa/index.md
+++ b/content/news/2020-07-30-galaxy-africa/index.md
@@ -1,7 +1,7 @@
 ---
 title: Galaxy Africa Community Update
 date: '2020-07-30'
-tags: [training, TIaaS, GCC]
+tags: [training, tiaas, GCC]
 location:
   name: online
 supporters:
--- a/content/news/2018-10-31-tiaas-feedback/index.md
+++ b/content/news/2018-10-31-tiaas-feedback/index.md
@@ -1,7 +1,7 @@
 ---
 title: Training Infrastructure Feedback from Maria Doyle
 date: '2018-10-31'
-tags: [training, TIaaS]
+tags: [training, tiaas]
 supporters:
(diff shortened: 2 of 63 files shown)


content/news/2020-07-30-galaxy-africa/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Galaxy Africa Community Update
 date: '2020-07-30'
-tags: [training, TIaaS, GCC]
+tags: [training, tiaas, GCC]
```
bgruening, line 4: Is GCC findable, or COVID-19?

dadrasarmin: There are many things that are not findable. For example, `single-cell` is not findable and there is an issue ( https://github.com/galaxyproject/galaxy-hub/issues/2200 ) for it on this repo. I need to check them one by one to see how they work. There is another one about https://github.com/galaxyproject/galaxy-hub/issues/2864 and another one here: https://github.com/galaxyproject/galaxy-hub/issues/407

Do you know an automatic way to solve these? My idea is to go through all tags that I once screened and unify them manually.
All tags were once screened locally by me here: 
https://github.com/galaxyproject/galaxy-hub/issues/3001#issuecomment-2754565178

PR #3175, opened 2025-06-11
update node to 18

martenson:
I assume this is not enough/proper. 

This PR is in a naive attempt to fix publish workflow, throwing:

```
error linkinator@6.1.2: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.20.2"
error Found incompatible module.
```

xref: https://github.com/galaxyproject/galaxy-hub/actions/runs/15577800186/job/43865985577

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v4
         with:
-          node-version: '16'
+          node-version: '18'
           cache: 'yarn'
       - name: Install project dependencies
         run: yarn --prefer-offline


bgruening:
@dannon pushing the artifacts is currently broken, do you think is PR is ok do go? Or will this cause major other problems?

dannon:
Argh, thanks for the ping, I hadn't seen this failure.  It wouldn't hurt to try, though node 18 is where all the webpack/ssl issues start happening in various configurations and there may be fallout we have to adjust for.

dannon:
Looks like it worked, thank you for noticing and fixing this!

PR #3179, opened 2025-06-11
News astro text analysis tool

Andrei-EPFL:
this a tool that has to be published soon on galaxy, but there are some checks that do not pass.

See https://github.com/esg-epfl-apc/tools-astro/pull/113


bgruening:
@Andrei-EPFL don't forget to add it to https://github.com/usegalaxy-eu/usegalaxy-eu-tools

PR #3181, opened 2025-06-12
Add DH SIG

Sch-Da:
Add DH SIG panel and link to index.

Is it working like this or am I missing something?

### DIFF
--- a/content/community/sig/index.md
+++ b/content/community/sig/index.md
@@ -110,7 +110,7 @@ Announcements and discussion in the Galaxy Public Health Community.
 <!-- Digital Humanities -->
 <div class="card" style="min-width: 12rem; max-width: 20rem">
 <a href="/community/sig/digital-humanities/">
-<div class="card-header bg-sig-field text-white text-center"><i class="fa fa-external-link-alt" aria-hidden="true"></i>&nbsp;Public Health</div>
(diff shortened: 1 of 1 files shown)


content/community/sig/index.md
```diff
@@ -107,6 +107,16 @@ Announcements and discussion in the Galaxy Public Health Community.
 
 </div>
 
+<!-- Digital Humanities -->
+<div class="card" style="min-width: 12rem; max-width: 20rem">
+<a href="/community/sig/digital-humanities/">
+<div class="card-header bg-sig-field text-white text-center"><i class="fa fa-external-link-alt" aria-hidden="true"></i>&nbsp;Public Health</div>
```
SaimMomin12: Digital humanities instead of Public Health

Sch-Da:
```suggestion
<div class="card-header bg-sig-field text-white text-center"><i class="fa fa-external-link-alt" aria-hidden="true"></i>&nbsp;Digital Humanties</div>
```

Sch-Da: yes, adapted, thanks for catching!

dadrasarmin: Now, that you mentioned it, I would like to highlight that nothing happens when I click on Public Health icon. For the rest, I will be redirected to their page.

Sch-Da: There was no link embedded for public health, which would explain this. Not sure, it is intentional, though

PR #3194, opened 2025-06-16
Wrote an event post for the tool wrapping workshop.

dadrasarmin:
This is the announcement for the tool wrapping workshop the European Galaxy team are going to organize in Freiburg. This will be the 1st session of hopefully many session we want to organize in future. Although the 1st session is an invitation only event, we would like to record it for future reference.


hechth:
@dadrasarmin I'd also be curious about the instructors ;) Also, will the materials from the workshop be available somewhere?

Issue #3135, opened 2025-05-13
Filtering based on tag metadata returns unexpected results on News section


dadrasarmin:
@ElectronicBlueberry has recently had[ a pull request ](https://github.com/galaxyproject/galaxy-hub/pull/3130)so we can search and filter for tags on News page. It works for most things well. For example, the following URLs work as expected:
1. https://galaxyproject.org/news/?tag=paper
2. https://galaxyproject.org/news/?tag=talk

It behaves strange on the hosted website for a few tags such as `TIaaS`. It works for the subsite: https://galaxyproject.org/projects/esg/news/?tag=TIaaS . However, it does not work properly on the global site: https://galaxyproject.org/news/?tag=TIaaS
Interestingly, it filters for posts that have other tags (for example, [this](https://galaxyproject.org/news/2025-04-26-bellavista-it/), [this](https://galaxyproject.org/news/2025-04-17-gat-survey/), and [this](https://galaxyproject.org/news/2025-04-07-galaxy-job-radar/)). One guess is that this happens because of the upper-case tag. However, it works well for the `esg` subsite.

Does someone have a suggestion on how to solve this?

Besides, if someone wants to look for a tag such as `paper` in subsites like eu or freiburg, they have do it as `https://galaxyproject.org/eu/news/?tag=paper` or `https://galaxyproject.org/freiburg/news/?tag=paper`. Ideally, we would like to be able filter for `tags` in all sub-sites when we are filtering globally `https://galaxyproject.org/news/?tag=something`. At the moment, this does not work globally, if the post does not have the `global` or `all` metadata for subsites.

dadrasarmin:
This should be solved with this: https://github.com/galaxyproject/galaxy-hub/pull/3161

Issue #1537, opened 2022-07-15
Allow filtering `taggedEventsPages` by subsite


NickSto:
Currently you can easily create a page listing events with a certain tag with the `taggedEventsPages` key in `config.json`. But if you want only those events relevant to a certain subsite, there's not a way to specify that.

We could solve that by adding an optional `subsite` key to each entry.

Issue #1099, opened 2021-12-22
List of all published Galactic News


beatrizserrano:
@erxleben is looking at the past Galactic News, and we can't find a list that gathers all the published ones. Does this exist?

We think that it would be useful to have something similar to this: https://galaxyproject.org/community/devroundtable/ 
I guess that a new section under Community (https://galaxyproject.org/community/newsletter/) would work. It would be just a filtered version of the News, and we could add general information, like how to subscribe, the frequency, how to contribute content (linking to the new form), etc.

beatrizserrano:
oh I just discovered this: https://galaxyproject.org/galaxy-updates/

can be that this has been manually updated?

NickSto:
Yeah that page isn't automated right now. I remember mentioning the idea of automating it to Dave but he didn't seem to think it was a high priority?

But it's not Dave's job anymore so if you guys would like an automated page then let's do it.

beatrizserrano:
Not high priority I think, but something to keep in mind :)

Issue #407, opened 2018-05-22
Filter events


bebatut:
Hi,

Would it be possible to filter events on the Hub, e.g. to get all events named "GTN contribution fest" or at a given date?
We discussed quickly with @tnabtaf about that.

Bérénice

PR #3198, opened 2025-06-18
added a section to search news and events using metadata.

dadrasarmin:
I find it sad that we cannot search news and events using metadata that we provide for each post. Now, it is a demonstration that we can search posts using title, dates, tags, supporters, subsites, and authors.

Of course, it can improve but I think having something to do some queries is nice. It is also asked in different issues before such as these:

- https://github.com/galaxyproject/galaxy-hub/issues/407
- https://github.com/galaxyproject/galaxy-hub/issues/1099
- https://github.com/galaxyproject/galaxy-hub/issues/2200
- https://github.com/galaxyproject/galaxy-hub/issues/2864
- https://github.com/galaxyproject/galaxy-hub/issues/3135
- https://github.com/galaxyproject/galaxy-hub/issues/1537
- https://github.com/galaxyproject/galaxy-hub/issues/252

We do not have to include it anywhere in the NavBar. I find its current place in the NavBar ugly. But I wanted to check whether there is any interest in such a thing and then I can improve the aesthetic.


src/components/NavBar.vue
```diff
@@ -23,6 +23,8 @@
                     <NavBarItem v-for="item of customContent.left" :key="item.key" :item="item" />
                 </b-navbar-nav>
                 <b-navbar-nav id="global-items" class="ml-auto">
+                    <!-- our new search-metadata link -->
+                    <b-nav-item to="/Search2" exact-active-class="active"> Meta-Search </b-nav-item>
```
arash77: I think this should help the lint error:
```suggestion
                    <b-nav-item to="/search2" exact-active-class="active"> Meta-Search </b-nav-item>
```

dadrasarmin: Thanks @arash77 ❤️

dadrasarmin:
The current lint error is not related to what I suggested in this PR and is mentioned previously in this issue: https://github.com/galaxyproject/galaxy-hub/issues/387 and it is about an internal page that does not exist now but there is a link to it here: https://github.com/galaxyproject/galaxy-hub/blob/e3eba062c97d2af0a0a3d058be60ebe8ad6d529d/content/learn/advanced-workflow/basic-editing/index.md?plain=1#L124

bgruening:
@dannon are you fine with it? It will not be visible for now, so we can try it in production and enable it in the main pages if it turn out to be useful.

dannon:
@bgruening I've been wanting to take a look at this but have not had time yet with GCC coming up.  I am not opposed to improving it of course, but what has to be kept in mind is that this is a 100% prebuilt static site -- there are no real dynamic queries here.  So any query is done clientside against some dataset, which is something prebuilt and shipped over the wire.  I would want to see how this affected the payload, how large an index we expect at 1000 posts, etc.

That said I don't mind at all this being (perhaps temporarily) in a second 'advanced news and events search' type page, but at the top level I am hesitant to add an extra search to the site -- I think there is significant value in having a single search that *just works* regardless of what you put in it.

dadrasarmin:
Thank you for your positive and honest feedback. I spent some time trying to reduce the size of the index and refactor the code, but haven’t landed on a sustainable solution that fully addresses the concerns raised. Because this PR doesn’t yet solve the core issues, I’m going to close it for now.

I’ve enjoyed working on this and appreciate how supportive this community has been. I look forward to contributing again in the future. Thank you.

dannon:
Hey, I just wanted to follow up and say thanks for jumping in and trying to build this out and enhance things!  It was a cool idea, but the static site performance stuff is just one of those annoying constraints we have to deal with.  Either way I'm glad you enjoyed working on this!

PR #3208, opened 2025-06-20
Adjust release 25.0 news item

ahmedhamidawan:
There were some incorrect elements there such as tool search improvements (which afaik weren't made in this release).


bgruening (approved):
If you thibk it's useful just use external_url and get people to your user release notes.

bgruening:
Thanks for fixing!

ahmedhamidawan:
> If you thibk it's useful just use external_url and get people to your user release notes.

Hmm, that's actually a cool idea.

But lets keep it like this for this release, and try that out next release onwards (no need to write a news item and just use an external url linking to release notes from 25.1 onwards)?

PR #3239, opened 2025-07-03
Update slug generation to use entry ID

arash77:
Change slug generation to utilize the entry ID instead of the link for improved consistency.
This should fix the folder name problem in external events like this https://github.com/galaxyproject/galaxy-hub/pull/3235


arash77:
For external URL posts like this, we will have a different structure that will lead to failure on some posts.
link->'https://forschungsdaten-thueringen.de/veranstaltung/Galaxy.html'
id->'https://training.galaxyproject.org/training-material/events/2025-11-26-TKFDM.html'

PR #3247, opened 2025-07-11
add presentation for EGD 2025

hrhotz:


hrhotz:
Lucille will register as soon as she has the 'approval' to come to Freiburg from her bosses. I have just put her name as a place holder so far

PR #3248, opened 2025-07-11
Propose a pyGenomeTracks talk

lldelisle:
Hi,
I propose to give a talk about pyGenomeTracks. The talk is very similar to what I have given in 2022 but maybe new people would be interested.


bgruening:
I like that. Can I maybe ask you to also include maybe 5 min about the new Visualization framework as well if I don't find anyone else :)

PR #3270, opened 2025-07-31
Wrote a post based on the WorkflowHub paper for the news Section (EuroScienceGateway).

dadrasarmin:
Hi,

I prepared a news item for the publication of WorkflowHub paper. The text has been prepared by the help of LLM but I have read and reviewed it.

@elichad @bgruening Could you please review the news item once before I publish it?


elichad:
Ah, I hadn't seen https://github.com/galaxyproject/training-material/pull/6259 - that actually covers nicely what I would want to highlight. 

I have added a review there and suggest we focus just on that one (though I took a couple of things I liked from this PR to suggest an "About WorkflowHub" section). It is a bit more focused on the specific integrations between WFH and GTN/Galaxy, which is what I wanted to aim for as I think that is the most interesting part for users.

PR #3283, opened 2025-08-13
Draft Belgium as dot star server

B0r1sD:
Hey all!

The useGalaxy Belgium team would like to open the discussion on the requirements on becoming an official usegalaxy.* or 'dot star' instance. 

Below I copied the current listed requirements (updated on [Dec 23, 2023](https://github.com/galaxyproject/galaxy-hub/pull/2262)) with a comment underneath where I added the current state of https://usegalaxy.be. Points of discussion marked in **bold**. 

### .* Membership current requirements

> *Updated*: These servers commit to updating Galaxy within 90 business days after new releases.

Since upgrading our Galaxy BE from version 24.1 to version 24.2, we've implemented a 'Request for change' or RFC method as part of change management that we are now applying to basically all our services. This solidifies the whole upgrade process as it comes with risk assessment, peer and technical review, step-by-step documentation, an upgrade checklist,... . We have been committed and successful in upgrading within 90 bussiness days, which we are now applying to v25.0. 

> *Freely accessible*: these servers may not place restrictions on access or registration, anonymous usage must be permitted.

We don't have any restrictions on access or registration as anyone can make an account. We do not have anonymous usage enabled, **but this might not be a hard requirement anymore?** 

> *Public infrastructure configuration*: we hold ourselves to a very high standard. Everyone should be able to reproduce what we do, and we commit ourselves to securing our infrastructure using best practices and in the eye of the public.

Most of the repositories of our usegalaxy-be are public on Github: https://github.com/orgs/usegalaxy-be/repositories?q=visibility%3Apublic. Our infrastructure playbook is not. **Is this still a hard requirement?** 

> *Common Tools & Reference Data*: once work in progress finishes, these servers will share a common set of tools, and must provide CVMFS reference data.

We provide CVMFS reference data (https://usegalaxy.be/libraries), and have 1400+ tools available on our instance. 

> *Functional*: the site provides >90% uptime as monitored by the [status page](https://status.galaxyproject.org/), and participates in automated cross-usegalaxy testing initiatives.

EDIT: We are now also part of [the official status page](https://status.galaxyproject.org/), and are also in [this one](https://stats.uptimerobot.com/gXd4fsKqVy). We have also been monitoring https://usegalaxy.be internally for some months, and can quite confidently say that we have been providing (and will continue to provide) well above 90 percent of uptime. We just lack the historical numbers to back this up properly. **What are these automated cross-usegalaxy testing initiatives refering to?**

PR #2657, opened 2024-06-23
dot-star updates following discussion

hexylena:
cc @frederikcoppens 

Other open tasks/issues

- [ ] Regions dropdown in the masthead should be removed (@nekrut)
- [x] Synchronise /use/ with /usegalaxy/ https://github.com/galaxyproject/galaxy-hub/commit/0d88ce0191c4f97967cd39c02521e1db0498c063
- [ ] galaxyproject.org "Use galaxy now" isn't synchronised
- [ ] add a google form
- [x] Add metrics https://github.com/galaxyproject/galaxy-hub/pull/2658
- [ ] specify the CVMFS.
- [ ] Google form (+ statement of how you'll contribute to the community)
- [ ] Instances should contribute the 'application's tab
- [ ] Minimum storage, 250GB

Related tasks: https://github.com/galaxyproject/galaxy/issues/13339

PR #3305, opened 2025-08-27
Update usegalaxy.* inclusion criteria

natefoo:
In light of the usegalaxy.be request at #3283, I propose the following updates to the inclusion criteria. The key change is to anonymous access: because funding often does not allow for anonymous jobs, I propose we allow inclusion as long as registration remains freely available and login is not required to access Galaxy, but may be required in order to run jobs.

That said, even servers that don't fully allow anonymous jobs I would encourage to allow some basic usage if possible: data source, text manipulation, fasta/fastq, etc. But I understand that for some it's not possible to allow any whatsoever, and don't think it should preclude membership in usegalaxy.*.

ping @B0r1sD @frederikcoppens @ccoulombe @carolg @abretaud @bgruening @cat-bro @GarethPrice-Aus @mschatz @nekrut

### DIFF
--- a/content/usegalaxy/index.md
+++ b/content/usegalaxy/index.md
@@ -43,10 +43,11 @@ This effort was launched in March 2018 and is very much a work in progress.  Sup
 We want to see servers meeting several loose requirements to be considered part of the `usegalaxy.*` group.
 
 - *Updated*: These servers commit to updating Galaxy within 90 business days after new releases.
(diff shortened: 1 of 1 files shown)


scottcain (commented):
👍

martenson:
this seems close to be a duplicate of #2657

jdavcs (approved):

PR #3348, opened 2025-09-19
Testimonials

dannon:
First pass at a quick component for this.  Pulls data from a single .yaml file, presents as something like this:

image

@scottcain Welcome any adjustments/suggestions/whatever.

### DIFF
--- a/content/data/testimonials.yml
+++ b/content/data/testimonials.yml
@@ -1,466 +1,466 @@
 heading: "What Galaxy Users Say"
 
 testimonials:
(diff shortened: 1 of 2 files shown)


scottcain:
@dannon This is great--thank you. I have a few questions related to the number of testimonials I added:

- Would it be possible to start the page at a random position in the array? Since there are so many, if we always start at zero, some would likely never get displayed.
- About the widget that shows progression when you mouse over it--what do you think about either shrinking it or removing it, since, on my screen, it runs off the end of the window?

dannon:
Hey, of course -- I'll have a random offset.  Happy to just drop the progression bit, too -- these aren't supposed to be clicked through I guess.  We could have a single page we link to that lists them all somewhere if we did want them in a more 'browsable' form.

scottcain:
@dannon I think this is ready to go! Thanks for your work on this

PR #3377, opened 2025-10-06
Update charts description and link

guerler:
This PR updates the Charts/Visualization Framework description and removes Trackster details (now that we have IGV), also removes Phyloviz (has been replaced with Phylocanvas).

screencast


bgruening:
What do you think about mentioning the other "vis" options here very briefly as well. Vis as tools, and Vis as ITs.
Should we also link to the GTN material, or is that deprecated?

guerler:
I added some example images to the document.

bgruening:
Thanks!

PR #3379, opened 2025-10-06
add ics muni as supporter

martenson:


content/images/logos/ics-muni.png
mvdbeek, line 1: @martenson I think the png is broken, I see this on the hub:
Screenshot 2025-10-06 at 14 08 56

martenson:
fixed

PR #3390, opened 2025-10-10
Add CloudFront rewrites function, test, and deployment workflow

natefoo:
The hub is currently served via nginx on an EC2 instance proxying the Hub S3 bucket. I'm moving this to CloudFront but needed a way to implement the [nginx-based rewrites](https://github.com/galaxyproject/infrastructure-playbook/blob/7a1d7068c21f370f39296f3cef2ebecc6b9e58eb/templates/nginx/galaxyproject.j2), which is now done via this function. I only added a couple testing rewrites for now and will follow up with the full set if everything works.

### DIFF
--- a/.github/workflows/rewrites.yml
+++ b/.github/workflows/rewrites.yml
@@ -1,0 +1,52 @@
+name: Deploy CloudFront Rewrites Function
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'deploy/rewrites.js'
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'deploy/rewrites.js'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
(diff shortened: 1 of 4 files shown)


.github/workflows/rewrites.yml
```diff
@@ -0,0 +1,52 @@
...
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
jplock, line 43: any chance you can update this to use GitHub OIDC instead of maintaining these long lived keys?

https://github.com/aws-actions/configure-aws-credentials?tab=readme-ov-file#oidc-configuration :)

natefoo: Well hey Justin, nice to see you here. :exploding_head: :smile:

PR #3430, opened 2025-10-27
Enhance feed checking logic in publish workflow

arash77:
The JSON files aren't updated with the new post upon publishing, so we have to wait and check when they get available.
This PR improve the validation process for new content availability in the publish workflow by implementing a more robust feed checking mechanism. This change allows for better handling of content updates and ensures that the workflow waits for the necessary feeds to be available before send it to galaxy social.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -46,14 +46,58 @@ jobs:
         uses: dorny/paths-filter@v3
         id: check_changes
         with:
+          list-files: json
           filters: |
             content:
-              - 'content/news/**'
-              - 'content/events/**'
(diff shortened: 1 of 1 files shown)


dannon:
Thanks for tracking this down! The issue is CloudFront serving cached versions
of feed.json even after we've uploaded fresh content to S3. Your polling
approach works around this by waiting for the cache to naturally refresh.

I'm going to follow up with a CloudFront cache invalidation step that will
proactively clear the cache for these feed files right after upload. Should
make this faster and more reliable.

arash77:
@dannon Thank you!

PR #3449, opened 2025-11-11
Add CloudFront cache invalidation for feed files

dannon:
Replace polling logic with proactive CloudFront cache invalidation. Instead of waiting up to 1 hour for feeds to naturally propagate, we now invalidate the cache for feed.json files after S3 sync and wait 90 seconds for the invalidation to complete.

Changes:
- Move content change detection to publish job
- Add AWS CLI cloudfront invalidation for /events/feed.json and /news/feed.json
- Replace Python polling script with simple cache invalidation
- Simplify galaxy-social-assistant trigger to only run when new content exists
- Pass days: "1" to social bot since we just published

This builds on top of PR #3430 which identified the CloudFront caching issue.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,9 +13,20 @@ env:
 jobs:
   publish:
     runs-on: ubuntu-latest
+    outputs:
+      has_new_content: ${{ steps.check_changes.outputs.content }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Check for news/events content changes
+        uses: dorny/paths-filter@v3
+        id: check_changes
+        with:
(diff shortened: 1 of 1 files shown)


arash77:
@dannon Thank you again for taking care of this.
It seems that [jakejarvis/s3-sync-action](https://github.com/jakejarvis/s3-sync-action) action has been archived. Can't we use something else, like the official one [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials) and use a normal `aws s3 sync` command after that?
Also, that would be great to update to `actions/checkout@v5` and `actions/setup-node@v6` with newer `node-version` like 20 or 24.

PR #3466, opened 2025-11-17
Fix/feed days ago filter

dannon:
Fix feed filtering for article with days_ago=0 -- classic error where when days_ago was 0, that was false, instead of null, but both caused the event to not be included.


dannon:
@arash77 this is my best guess of what was going on with the feed issues -- let me know if this helps.  It definitely fixes at least one bug :)

arash77:
Thank you @dannon.

PR #3499, opened 2025-12-17
Fix Node 22 url.parse() error on navbar keys

nekrut:
Change navbar item key separator from : to | to avoid ERR_INVALID_ARG_VALUE when Node 22's stricter url.parse() encounters keys like "link:subsite:news/"


hujambo-dunia:
At any rate, I prefer to avoid merging on Friday's. So I'll take the addition time to take a closer look there might be a more sustainable way to accomplish this. Thank you for bringing this to my attention.

hujambo-dunia (approved):

PR #3547, opened 2026-01-15
Fix internal links for astro-preview subpath deployment

dannon:
## Summary
- Added an Astro integration that rewrites internal `href` links to include the base URL prefix after build
- Only activates when `ASTRO_BASE` env var is set (e.g., `/astro-preview/`)
- Local dev and root deployments are unaffected

Fixes the issue where clicking links at https://galaxyproject.org/astro-preview/ would navigate back to the old Gridsome site instead of staying within the Astro preview.


bgruening:
Awesom

PR #3573, opened 2026-01-17
Add support for future-dated news articles

### LINKED ISSUE #3565: Add support for "future-posts"
Add support for “future-news”. News have dates in the filename. When this date is in the future the news item should __not__ be rendered yet on the page, but rendered when the date is reached and the site is rebuild.

This will enable us to prepare news items, e.g. for GCC etc in advance.

dannon:
Closes #3565

## Summary

Adds filtering for future-dated news articles at build time. This allows preparing news content in advance (e.g., for GCC announcements) without it appearing on the site until the specified date.

- Added `getPublishedNews()` helper that filters out articles with future dates
- Updated news listing pages, individual article generation, and RSS feed to use this helper
- Added unit tests for the date filtering logic

## Note on publishing

Since this is a static site, future-dated articles will only become visible after a site rebuild on or after their publication date. Post-Astro cutover, we plan to add a nightly rebuild job to automatically publish scheduled content.

### DIFF
--- a/.github/workflows/astro-checks.yml
+++ b/.github/workflows/astro-checks.yml
@@ -55,6 +55,12 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Preprocess content
+        run: npm run preprocess
+
+      - name: Generate search index
+        run: npm run search-index
+
       - name: Install Playwright browsers
         run: npx playwright install --with-deps chromium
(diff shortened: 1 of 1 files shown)


dannon:
Also fixes a minor bug in CI when using the lighter weight build for playwright.

Issue #3569, opened 2026-01-17
sub-headers should be linkable


bgruening:
All headers and sections on pages should have an anchor that is reference-able / link-able

e.g. those at https://preview.galaxyproject.org/publication-library/

<img width="712" height="260" alt="Image" src="https://github.com/user-attachments/assets/a71947ae-1ba0-48e4-95ee-5cff6ba9122f" />

PR #3578, opened 2026-01-18
make all headers linkable in astro

bgruening:
fixes: https://github.com/galaxyproject/galaxy-hub/issues/3569


dannon:
With the merge of #3591 we should be able to close this now.

PR #3581, opened 2026-01-18
[astro] add citation subsite

bgruening:
This adds a similar page as https://usegalaxy-eu.github.io/citations to the new astro based website.

The URL is /eu/citations or any subsite. In theory, we can fetch those citations directly from [Zotero](https://www.zotero.org/groups/1732893/galaxy/), but this would increase the build-time. So I decided (like on the old page) to have a different CI jobs that populates the BibTeX file and astro is just parsing the BibTeX entries. For this I need a new dependency `citations-js`.

fixes: https://github.com/galaxyproject/galaxy-hub/issues/3582


content/eu/citations/citations-eu.bib
bgruening, line 1: @dannon if this PR gets accepted I will add a CI that updates it from https://www.zotero.org/groups/1732893/galaxy/ 

xref: https://github.com/usegalaxy-eu/website/blob/master/.github/workflows/update-citations.yml

Issue #3584, opened 2026-01-18
Make the year-filters on the news and events sites routable


bgruening:
<img width="411" height="334" alt="Image" src="https://github.com/user-attachments/assets/289c082e-1d95-479e-8cac-bc1c02868efe" />

Addressing certain years should be possible.

PR #3585, opened 2026-01-18
Make the years-filter in news and event pages routable

bgruening:
fixes: https://github.com/galaxyproject/galaxy-hub/issues/3584


astro/tests/year-routing.spec.ts
bgruening, line 1: Are those tests needed? Do you want them somewhere else included? Or do we add them for now and restructure them later?

dannon: It's good to add them like this for now, very helpful!

Issue #3589, opened 2026-01-18
/use/ needs to render strings in markdown


bgruening:
On the /use/ page and subpages we need markdown rendering, currently its rendered as text.

<img width="957" height="214" alt="Image" src="https://github.com/user-attachments/assets/846f9116-3263-41de-93e9-322eae1b1ef4" />

PR #3590, opened 2026-01-18
Fix rendering on /use/ and /use/<server>

bgruening:
fixes https://github.com/galaxyproject/galaxy-hub/issues/3589

### DIFF
--- a/astro/src/layouts/PlatformLayout.astro
+++ b/astro/src/layouts/PlatformLayout.astro
@@ -64,12 +64,7 @@ const scopeNames: Record<string, string> = {
   'tool-publishing': 'Tool Publishing',
 };
 
-const renderMarkdown = (value?: string) => (value ? marked.parse(value) : '');
-const renderListItemMarkdown = (value?: string) => {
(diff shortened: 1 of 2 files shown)


astro/src/layouts/PlatformLayout.astro
```diff
@@ -63,6 +64,13 @@ const scopeNames: Record<string, string> = {
   'tool-publishing': 'Tool Publishing',
 };
 
+const renderMarkdown = (value?: string) => (value ? marked.parse(value) : '');
+const renderListItemMarkdown = (value?: string) => {
+  if (!value) return '';
+  const html = marked.parse(value).trim();
+  return html.startsWith('<p>') && html.endsWith('</p>') ? html.slice(3, -4) : html;
```
bgruening: that hack is needed, as otherwise the list is rendered as `<li><p>text</p></li>` which includes a line-break after the bulletpoint :(

dannon: I think the commit I pushed should address this.  Thank you!

PR #3591, opened 2026-01-19
Add table of contents sidebar and linkable headers

### LINKED ISSUE #3569: sub-headers should be linkable

dannon:
## Summary
- Adds automatic table of contents (TOC) sidebar for article pages, matching the old Gridsome behavior
- Adds clickable anchor links to all headings for easy section linking

## Features

**Table of Contents:**
- Displays on the right sidebar (desktop) or as a collapsible section (mobile)
- Shows h2-h4 headings with proper indentation
- Only appears when page has 2+ headings
- Controlled via `autotoc` frontmatter field (opt-in, matching old behavior)

**Linkable Headers:**
- All headings now have anchor links that appear on hover
- Clicking copies the section URL to the address bar
- Uses `#` symbol, styled to be unobtrusive

## Technical Changes
- Added `rehype-slug` and `rehype-autolink-headings` plugins
- Created `TableOfContents.astro` component
- Updated `ArticleLayout.astro` and event pages to support TOC
- Page templates now extract and pass headings from rendered content

Fixes #3569

### DIFF
--- a/astro/src/components/layout/ContentWithToc.astro
+++ b/astro/src/components/layout/ContentWithToc.astro
@@ -1,0 +1,59 @@
+---
+/**
+ * ContentWithToc - Wraps article content with optional Table of Contents
+ *
+ * Handles both mobile (collapsible) and desktop (sidebar) TOC display.
+ * Use this component to avoid duplicating TOC logic across page templates.
+ */
+import TableOfContents from './TableOfContents.astro';
+
+interface Heading {
+  depth: number;
+  slug: string;
+  text: string;
+}
+
+interface Props {
+  headings?: Heading[];
+  autotoc?: boolean;
+}
+
+const { headings = [], autotoc = false } = Astro.props;
+
+// Determine if TOC should be shown (autotoc enabled and enough headings)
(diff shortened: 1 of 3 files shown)


astro/astro.config.mjs
```diff
@@ -4,11 +4,30 @@ import vue from '@astrojs/vue';
 import tailwindcss from '@tailwindcss/vite';
 import mdx from '@astrojs/mdx';
 import sitemap from '@astrojs/sitemap';
+import rehypeSlug from 'rehype-slug';
+import rehypeAutolinkHeadings from 'rehype-autolink-headings';
+
+const autolinkConfig = {
+  behavior: 'prepend',
```
bgruening: Imho, this does not look good. All headers have an indentation that looks off. ~~Unfortunately also the `append` looks not good as it is on the far right of the site.~~

Playing around with it, I think `append` could work. I would prefer that over `prepand`

PR #3599, opened 2026-01-20
Supporters and tags were added to multiple news and event items. Part 1

dadrasarmin:


astro/package-lock.json
bgruening, line 1: that should maybe not be removed.

content/news/2020-09-eosc-calls/index.md
```diff
@@ -5,6 +5,8 @@ tease: Two funding opportunities are now open in EOSC-Life.
 tags: [grants]
 supporters:
 - eosc
+- EOSC-Life
```
bgruening, line 8: do we want to standardise that to all lower-letter?

bgruening: what is GTN doing?

content/news/2021-08-elixir-update/index.md
```diff
@@ -4,7 +4,8 @@ date: '2021-08-24'
 tease: An updated ELIXIR Galaxy website includes our goals, with a special emphasis
   on interactions with the local, domain-specific and global communities; plus much
   more
-tags: [COVID-19, training, galaxy]
+tags: [COVID-19, training, galaxy, elixir, community]
```
bgruening:
```suggestion
tags: [covid-19, training, galaxy, elixir, community]
```

I have seen occurrences with just small letters

content/news/2020-09-eosc-calls/index.md
dadrasarmin: We can. But we can do that later as well. Because this is not the only case. There are many different variations of upper, lower, and mix typesetting of different supporters: https://github.com/galaxyproject/galaxy-hub/blob/6fb51b6f333c7e922ddd4de4caf72a558abd44bd/content/supporters.yml

I suggest we solve that in another PR and for all of them. Better if we can do it via a script both in the supporters.yaml file and in each news item that they appear.

Issue #3595, opened 2026-01-19
Proportional font size of tags in Astro development is not practical and/or visually pleasing


dadrasarmin:
In the new website, there is a very useful section to filter based on Tags: http://localhost:4321/tags/
I noticed that the size of the font is somehow proportionate to the usage of each tag. However, since there are few (3? Tools, Training, DevOps) tags with around 100 usage, everything else is shown with almost the same font size.

<img width="1866" height="1283" alt="Image" src="https://github.com/user-attachments/assets/832940e6-6b61-47a4-bc0e-7da68a416508" />

I personally prefer the items in that page to not be proportional to the usage. However, if that is inevitable, I think there should be a normalization in place to make visual clues more meaningful.

dadrasarmin:
I think sorting in each letter is a good trade-off. Thanks.

PR #3604, opened 2026-01-21
astro: tags display adjustments

dannon:
closes https://github.com/galaxyproject/galaxy-hub/issues/3595


dadrasarmin:
If at some point, you want to include a hall of fame for Galaxy Hub (https://github.com/galaxyproject/galaxy-hub/pull/3574), the font size of authors' names and supporters are also proportional to the number of their appearance similar to tags. Since again we have a very few names with lots of contribution, the proportional size become almost meaningless unless we do a similar sorting there.

I guess the adjustment of the font size could be the same as tags but for `authors` and `supporters`.

PR #3614, opened 2026-01-22
Next iteration of the Hall of Fame

bgruening:
@dannon I hope this goes into the left direction.

One crucial thing is that we are now shipping the ORGANISATION.yaml, GRANTS.yaml and CONTRIBUTORS.yaml file + shemas here. We should add a CI jobs that syncs that with GTN. But for now I would like have it here to play around with all the data.

There is one post/news included in this PR, that has both (old and new) metadata schemas included. If both are included, the new one is preferred.

I have decided to NOT render twitter badges and accounts on purpose.

Here are a few screenshots:

image

------

image

------

image


content/news/2026-01-07-hf-integration/index.md
```diff
@@ -8,6 +8,8 @@ contributions:
   funding:
     - uni-freiburg
     - nfdi4bioimage
+supporters:
```
bgruening, line 11: @dadrasarmin that is only here for testing. I think we can merge, it does not matter, but it should test that supporters is rendered in the old side.

Where on the new site, IFF we do have `contributions:` the `supporters` are skipped. `supporters` are only used if there are no `contributions:` on the new site.

content/CONTRIBUTORS.yaml
bgruening, line 1: @dannon this is kind of an architectural decision. 
I think it's ok to ship those files and I would add a CI job that get them from GTN. For the time being we tell people to add there metadata in those files at GTN.

content/schema-contributors.yaml
bgruening, line 1: Those schema files are not strictly needed, but they might help LLMs and I think its a good documentation. I would update them as well with a CI.

astro/src/assets/gtn_icon.svg
bgruening, line 1: This is needed because astro can only ship stuff under src/ ?

content/news/2026-01-07-hf-integration/index.md
dadrasarmin: Very nice strategy.

content/CONTRIBUTORS.yaml
dannon: This sounds fine to me to ship them for now and sync eventually.

bgruening: Great, I think then this PR is ready from my side.

PR #3619, opened 2026-01-24
[astro] Add people-pages to sub-sites

bgruening:
This PR introduces a new people-page. The idea is to create for a team or a lab a space to present themselves.
It's inspired by the EU people site(s) that we never migrated to Gridsome: https://galaxyproject.eu/people

The people pages are not linked in the global menu, but as soon as you visit one of the sub-sites you will see:

grafik

We introduce here a PersonCard, that is reused on multiple sites:

grafik

grafik

The `Europe` site is a bit special, as it is an aggregate of all the European sub-sites, like France, Belgium, Freiburg, Italy etc ... The data is extracted from [EU](https://galaxyproject.eu/people) and overlaps a bit with the hall-of-fame metadata. I hope that is ok, for the time being, the People pages are meant for sub-sites to represent people working on Galaxy servers and to populate lab-sites.

On the individual sub-sites, NOT the EU-aggregate sites, there is also support for Alumni's ... so we can also list them. You can see that in action under /freiburg/people/

### DIFF
--- a/astro/tests/redirects.spec.ts
+++ b/astro/tests/redirects.spec.ts
@@ -2,7 +2,7 @@ import { test, expect } from '@playwright/test';
 
 test.describe('External URL redirects', () => {
   test('news item with external_url redirects to target', async ({ page }) => {
-    const response = await page.goto('/news/2024-12-19-community-page/');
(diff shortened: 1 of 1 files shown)


bgruening:
@dannon green now!

PR #3622, opened 2026-01-25
add initial schema validation

bgruening:
Initial attempt to add some validation to our news and event items.

Currently, this is the schema that we have currently present, with all old cruft, wrong subsites and redundant tags.

The idea would be to merge this and then change the schema files, see errors, adopt the content piece by piece.

A few comments below as review comments.

@dadrasarmin maybe that is useful for you.

It can already find news items with subsites "esg-wp1" ... so it might be useful, but we should make it more stringent.


content/SUBSITES.yaml
```diff
@@ -0,0 +1,26 @@
...
+- astronomy
+- astrophysics
+- belgium
+- collaboration
+- elixir-it
+- eosc
+- erasmusmc
+- esg
+- esg-wp1
+- esg-wp4
```
bgruening, line 13: Those should be removed I think. At the moment we have everything that I could find included here.

content/SUBSITES.yaml
```diff
@@ -0,0 +1,26 @@
...
+- eu
+- fr
+- freiburg
+- genouest
+- global
+- ifb
+- outreach
+- pasteur
+- pulsar
+- rocrate
```
bgruening, line 23:
```suggestion
```

content/TAGS.yaml
```diff
@@ -0,0 +1,257 @@
+---
```
bgruening, line 1: This list should be filtered imho, tags should be merged etc ... and if we see many tags like "Bielefeld", "Brno" maybe we should use `location`, which should exist, not sure if rendered in any of the both sites.

content/schema-events.yaml
```diff
@@ -0,0 +1,107 @@
...
+        location:
+            name: Online, Global
+mapping:
+    "=":
+        type: map
+        mapping:
+            title:
+                type: str
+                description: Event title
+            date:
```
bgruening, line 19: needs stricter validation at some point

content/schema-events.yaml
```diff
@@ -0,0 +1,107 @@
...
+                type: str
+            hide_tease:
+                type: bool
+            tags:
+                type: seq
+                sequence:
+                    - type: str
+                      enum:
+                          - TAGS
+            subsites:
```
bgruening, line 38: maybe at a description where the SUBSITES.yaml can be found

scripts/validate_common.py
```diff
@@ -0,0 +1,191 @@
...
+                    line = line[2:]
+                messages.append(line)
+            if messages:
+                break
+        if not messages:
+            messages.append(str(exc))
+        return 1, messages
+
+
+def check_recent_folder_names(aggregated, cutoff="2026-02-01"):
```
bgruening, line 147: The idea behind cutoff is that we don't want to change the location of old posts, we need to keep the URLs stable. However, it would be nice to validate folder names, and hence the URLs a bit. So wrong a given time-point onwards we validate folder names.

Makefile
```diff
@@ -56,6 +56,11 @@ astro-preview: astro-install ## Preview built Astro site
 	cd astro && npm run preview
 .PHONY: astro-preview
 
+validate-metadata: ## Validate news and events frontmatter schemas
+	python scripts/validate_news.py --be-strict-from 2026-02-01
+	python scripts/validate_events.py --be-strict-from 2026-02-01
```
bgruening, line 61: @dannon this date needs to be adjusted when we do the switch.

PR #3641, opened 2026-02-01
[astro] missing images in subdirs

bgruening:
Turns out we have been missing many images that were located in news/subdirs e.g. /news/2025.xxxx/images/

Adds a make target `astro-check` that lints, formats and checks the astro build.

The problem can be seen for example on this site: https://preview.galaxyproject.org/news/2025-12-05-de_biohackathon/ ... if you scroll down there are two images missing.


astro/astro.config.mjs
```diff
@@ -32,6 +32,16 @@ export default defineConfig({
...
   },
   vite: {
-    plugins: [tailwindcss()]
+    plugins: [tailwindcss()],
+    server: {
+      // Reduce inotify watcher pressure during dev/Playwright runs
+      watch: {
+        usePolling: true,
+        ignored: [
+          '**/public/images/**',
```
bgruening, line 41: Imho we could also unwatch all images, e.g. all jpg, png, etc ...

PR #3643, opened 2026-02-01
dealing-with-more-link issues

bgruening:
List of missing images: https://gist.github.com/bgruening/92f2565bd082ae79223d0a1e6b774fe8

### DIFF
--- a/astro/src/build/preprocess.mjs
+++ b/astro/src/build/preprocess.mjs
@@ -291,32 +291,31 @@ function slugToFilename(slug, useMdx = false) {
 }
 
 /**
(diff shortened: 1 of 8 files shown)


astro/src/build/preprocess.mjs
```diff
@@ -222,25 +222,37 @@ function slugToFilename(slug, useMdx = false) {
 function processImagePaths(content, slug) {
   let processed = content;
 
+  function rewriteSrc(src) {
```
bgruening: Maybe you have a better advice to tackle this issue.

We can add more rewrite options, or we just wait for the switch and fix those issues in the main source MD files.

PR #3660, opened 2026-02-05
Fix broken kramdown attributes in get-started and other pages

dannon:
## Summary
- Fixes broken table links on `/get-started/` where `{:target="_blank"}` was rendering as plain text
- Adds preprocessing step to convert kramdown-style attributes to proper HTML
- Affects 40 legacy content files from 2018-2021 that used Jekyll/kramdown syntax

## Details
Kramdown syntax like `[text](url){:target="_blank"}` is not supported by Astro's markdown processor. This was causing the tutorials table on the get-started page to display broken links with literal `{:target="_blank"}` text.

The fix converts these during preprocessing to proper HTML anchor tags:
- `[text](url){:target="_blank"}` → `<a href="url" target="_blank">text</a>`
- `[text](url){: .btn .btn-primary}` → `<a href="url" class="btn btn-primary">text</a>`

This approach fits with other legacy syntax conversions already in the preprocessing pipeline (g-link, Vue bindings, etc.).

## Test plan
- [x] Unit tests added (9 test cases)
- [x] All 62 unit tests pass
- [x] E2E tests pass (2 pre-existing flaky tests excluded)
- [x] Verified get-started page renders correctly in dev server

### DIFF
--- a/astro/tests/content-rendering.spec.ts
+++ b/astro/tests/content-rendering.spec.ts
@@ -1,22 +1,27 @@
 import { test, expect, type Page, type Locator } from '@playwright/test';
 
 /**
- * Click a link and wait for Astro's view transition to complete.
- * Astro fires 'astro:page-load' when the transition finishes and the new page is visible.
+ * Click a link and wait for navigation to complete.
+ * Works with both full-page navigations and Astro view transitions.
  */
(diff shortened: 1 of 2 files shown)


bgruening (approved):
I'm looking so forward to the switch and go to all sites and remove that for real :-)

PR #3663, opened 2026-02-05
[astro] Homepage content updates: new sections, updated stats and messaging

dannon:
## Summary

Reworks the Astro homepage with updated stats, new sections, and revised messaging that leans into Galaxy's core identity.

- **Stats row**: Updated from Anton's infographics — 400K+ registered users, 750K jobs/month, 22K+ citations, 150+ countries
- **Pillars**: Replaced feature-oriented pillars (Analyze/Reproduce/Learn/Scale) with value-oriented ones matching the classic Galaxy tagline: Accessible, Reproducible, Transparent. Section header is "Data intensive science for *everyone*"
- **How Galaxy Works**: 3-step flow (Bring Your Data → Run Your Analysis → Share & Reproduce) that better reflects how people actually use Galaxy (many run curated workflows, not just pick individual tools)
- **Impact Spotlight**: New section with VGP and Galaxy Training Network cards as concrete evidence of Galaxy's reach
- **Hero**: Revised tagline (removed redundant 400K stat that's now in the stats row), added server pill links for usegalaxy.org/eu/org.au
- **CTA**: Three server buttons with region labels instead of generic "Start Using Galaxy"

### Still up for discussion

- We may or may not want to keep both the pillars section and the 1-2-3 steps — there's some thematic overlap and the page might be tighter with just one of them. Open to feedback on this.
- The pillar cards (Accessible, Reproducible, Transparent) should eventually link somewhere — they look and feel clickable but don't have destinations yet. Not doing content work in this pass though.
- Copy and layout can still be iterated on, this is a solid starting point but not necessarily final.

### DIFF
--- a/content/people/people.yaml
+++ b/content/people/people.yaml
@@ -201,7 +201,7 @@ freiburg:
         website: http://leendertse.eu
         linkedin: JanLeendertse
         researchgate: JanLeendertse
-        matrix: @janleendertse:matrix.org
+        matrix: "@janleendertse:matrix.org"
 
     beatrizserrano:
         name: Beatriz Serrano-Solano


astro/src/pages/index.astro
wm75:
```suggestion
            <a href="https://mstdn.science/@galaxyproject" aria-label="Mastodon">
```

remaining link fixes in #3681

PR #3687, opened 2026-02-11
Serve legacy static assets in Astro build

dannon:
## Summary

- Extend preprocessing to copy `content/assets/` and `content/media/` into `astro/public/`, following the same pattern already used for `content/images/`. This makes ~145 `/assets/media/...` references across 88 content files and the `/media/...` references work in the Astro site.
- Stop tracking `generated-redirects.json` (build artifact that shouldn't be in version control).

## Test plan

- [x] `npm run preprocess` copies 198 asset files and 32 media files
- [x] `npm run build` succeeds (5370 pages)
- [x] `astro preview` serves all referenced URLs with 200 (`/assets/media/*.png`, `/media/freebayes.pdf`)
- [x] `npm run test:unit` — 93/93 passed
- [x] `npm run test` (Playwright) — 87/87 passed

### DIFF
--- a/content/people/people.yaml
+++ b/content/people/people.yaml
@@ -630,3 +630,22 @@ cz:
         galaxyadmin: true
         bio: |
           I'm an engineer focused on empowering scientists to perform reproducible data analysis.
+
+    demon:
+        name: Martin Demko
+        title: |
+            Dr. rer. nat., Ph.D. in Computer Science &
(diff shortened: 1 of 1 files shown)


bgruening:
I found the file useful, so see the mapping, but I guess its ok. Thanks @dannon

PR #3690, opened 2026-02-11
Search page styling and tab switch fix

dannon:
## Summary

- Use standard page container for search page (PageHeader + white card body), matching layout of news/events/etc
- Style search tabs with Galaxy-branded colors instead of default shadcn neutral
- Fix Pan-Galactic and Publication search not executing when switching tabs — `forceMount` keeps components alive so Google CSE's global state and Zotero results persist across tab switches

## Test plan

- [ ] Navigate to /search, type a query, confirm Hub Search works
- [ ] Click Pan-Galactic Search tab — Google results should appear
- [ ] Click Publication Search tab — Zotero results should appear
- [ ] Switch back to Hub, change query, switch to other tabs — results should update
- [ ] Load directly via URL: `/search?q=cloud&tab=google` — should show Google results
- [ ] Verify page layout matches other pages (news, events) with PageHeader and card body

### DIFF
--- a/astro/src/components/search/SearchPage.vue
+++ b/astro/src/components/search/SearchPage.vue
@@ -58,20 +58,35 @@ watch(activeTab, (tab) => {
     <!-- Tabs -->
     <Tabs v-model="activeTab" default-value="hub">
       <TabsList class="mb-6 bg-galaxy-dark/10 p-1">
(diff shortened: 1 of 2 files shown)


wm75:
@dannon that last commit is fixing the switch from searching in the side bar to the central search results?

PR #3712, opened 2026-02-17
[astro] Update US team details

ahmedhamidawan:
Based on the work done in #3619, I've added US team details based on information I could gather.

firefox_nYQm5HsGef

## PR Summary
- Expands the US Galaxy team `/us/people/` section by splitting the single `us` group into multiple distinct team
 subsites (like **Johns Hopkins**, **Penn State**, ...) _(please add more teams as needed)_
 - Adds base details (name, title, email, links, etc.) for people across several US teams, gathered from publicly
 available info — folks are encouraged to review and update their own entries
 - Refactors `europeSites` from a plain array to a label-mapped object (same pattern now used for the new
 `usSites`), so we can use the ids as keys to refer to the yml, and labels to display in the UI
 - Extracts the inline site-navigation links into a reusable `SiteNavigation` Vue component, now shared by both
 the EU and US people pages

### DIFF
--- a/astro/src/utils/people.ts
+++ b/astro/src/utils/people.ts
@@ -6,7 +6,7 @@ import { communitySlug } from './contributors';
 
 export const europeSites = {
   freiburg: 'Freiburg',
-  ifb: 'IFB',
+  ifb: 'ELIXIR France/IFB',
   erasmusmc: 'Erasmus MC',
   'elixir-it': 'ELIXIR Italy',
   genouest: 'GenOuest',
--- a/astro/tests/static-pages.spec.ts
+++ b/astro/tests/static-pages.spec.ts
@@ -188,15 +188,15 @@ test.describe('Static Pages', () => {
     test('/ifb/people/ page loads', async ({ page }) => {
       const response = await page.goto('/ifb/people/');
       expect(response?.status()).toBe(200);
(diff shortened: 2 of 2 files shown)


astro/src/utils/people.ts
```diff
@@ -4,9 +4,22 @@ import { fileURLToPath } from 'url';
 import { parse } from 'yaml';
 import { communitySlug } from './contributors';
 
-export const europeSites = ['freiburg', 'ifb', 'erasmusmc', 'elixir-it', 'genouest', 'cz'] as const;
-
-export type EuropeSiteId = (typeof europeSites)[number];
+export const europeSites = {
+  freiburg: 'Freiburg',
```
bgruening, line 8: Should we maybe move that also into a config YAML file somewhere? So that it is independent of the "implementation/framework"

ahmedhamidawan: I like that 👍 

Will do that

ahmedhamidawan:
> Maybe it's also nice to add previous people? We included the `alumni: true` feature for that.

Yes this is in fact very incomplete (see the [PSU team](https://preview.galaxyproject.org/us/people/#pennstate) for example). I added James to the JHU alumni but couldn't gather info on all other alumni from other teams. It'd be ideal if people from those teams can pitch in and either add those changes or let us know and I can make them. 

I was just afraid of adding any inaccurate info 😬

PR #3731, opened 2026-02-22
[astro] add sitemap for bare-pages

bgruening:
This is adding a sitemap to the /bare/ route. It will serve as a shop for people to know what they can embed, including a simple example.

grafik

@hujambo-dunia not all the bare-sites listed here are styled properly. Is that expected after your changes? The EU sites looks now beautifully, thanks a lot :)

### DIFF
--- a/astro/src/pages/bare/index.astro
+++ b/astro/src/pages/bare/index.astro
@@ -1,0 +1,115 @@
+---
+import { getCollection } from 'astro:content';
+import BareArticleLayout from '../../layouts/BareArticleLayout.astro';
+
+function normalizeRoute(path: string): string {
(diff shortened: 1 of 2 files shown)


hujambo-dunia:
> @hujambo-dunia not all the bare-sites listed here are styled properly. Is that expected after your changes? The EU sites looks now beautifully, thanks a lot :)

@bgruening Thanks - I had missed some of these. To clarify, the specific bare routes you believe still need updating/fixing are just these two pages only: `/bare/eu/usegalaxy/metabolomics/` and `/bare/eu/usegalaxy/proteomics/` as a part of production cutover? For example, then the `*/bare/eu/events` doesn't need further major changes? (Although I might modify the table header rows so they are all consistent across the bare pages.)

PR #3752, opened 2026-02-25
Fix duplicate content ID warnings from Astro glob-loader

dannon:
## Summary

- Delete stale `content/news/2022-02-01-27001-certifiation/` directory (typo copy of the correctly-spelled `certification` dir with older content)
- Add duplicate slug detection to the preprocessor so collisions are caught at preprocess time instead of surfacing as confusing Astro glob-loader warnings
- Add a regression test that runs the full preprocessing pipeline and asserts zero slug collisions

## Test plan

- `npm run preprocess` produces clean output with no duplicate warnings
- `npm run test:unit` passes all 154 tests including the new collision detection test


astro/src/build/preprocess.mjs
```diff
@@ -919,6 +919,41 @@ function convertVueToJsx(content) {
   return processed;
 }
 
+/**
+ * Extract a tease (short description) from raw markdown body text.
+ * Strips markdown/HTML syntax and returns the first sentence or ~200 chars,
```
bgruening, line 924: Not a fan of those modifications. If we decide to not want markup in the tease, we should remove it from the content (people look at older stuff and take inspiration), lint against it and fail any PR.

dannon: Yeah, this was just a shortcut for older content that didn't have any tease specified, but it's a good point that it could be seen as an example.  Happy to follow up and regenerate better ones later if we want, for sure!

PR #3757, opened 2026-02-25
Add microGalaxy roadmap and adapt page for astro

bebatut:


content/community/sig/microbial/index.md
```diff
@@ -12,7 +12,7 @@ title: "Microbiological Data Analysis in Galaxy"
 
 Your hub for **microbiological data analysis** in Galaxy, whether you're analyzing microbiome samples, bacterial isolates, long or short reads, shotgun or 16S data, genomics, transcriptomics, proteomics, metabolomics, or multi-omics integrative analyses.
 
-# Community Goals
+## Community Goals
```
bgruening, line 15: @dannon does this look ok to you? We should adopt this in astro's default style isn't it?

dannon: @bgruening Yeah, I'm working on a bit to automatically shift headings; contributors should ideally be able to write a coherent markdown document using h1s as desired and, depending on where it ends up being embedded, we automatically shift headings appropriately (shifitng h1->h2, etc., if it's on a page with a pre-existing title)

PR #3801, opened 2026-03-06
Edge redirect: normalize slugs in CloudFront function

dannon:
## Summary

Astro in static mode generates `<meta http-equiv="refresh">` HTML pages for redirects — so visiting a non-normalized URL like `/events/GCC2026/` causes the browser to load an HTML page, parse the meta tag, and then navigate. The user sees a visible redirect flash.

This PR moves slug normalization to the CloudFront Viewer Request function, which runs at the edge **before** S3 is hit. Non-normalized URLs now get an immediate HTTP 302 with no round-trip to S3.

## Changes

- **`deploy/rewrites.js`**: Added `normalizeSegment()` and `normalizeUri()` matching the rules in `slug-utils.mjs` — camelCase split, uppercase-run split, underscore→hyphen, lowercase, collapse hyphens, apply overrides. Date-prefixed segments (`YYYY-MM-DD-...`) pass through unchanged. If the normalized URI differs from the request URI, returns 302 immediately before checking the manual patterns list.

- **`deploy/test-rewrites.js`**: Added tests for camelCase redirect, underscore redirect, `gcc2026` passing through unchanged (no letter↔digit split), already-normalized slug passing through, and date-prefixed segment passing through. All 13 tests pass.

## Notes

- The normalization algorithm intentionally does **not** split letter↔digit boundaries (`gcc2026` stays `gcc2026`). This matches the current `slug-utils.mjs` behavior.
- The inline `slug-overrides` dict in `rewrites.js` must stay in sync with `slug-overrides.json` — both are updated here.
- CloudFront Functions have a 10KB size limit; this implementation is well within that.
- The meta-refresh HTML files remain in S3 as fallback but are never reached for unnormalized slugs since the CloudFront function intercepts first.
- Pairs with a separate PR that pre-bakes all known redirects as S3 object metadata for native 301s.

### DIFF
--- a/deploy/rewrites.js
+++ b/deploy/rewrites.js
@@ -1,6 +1,50 @@
+// Normalize a single URI segment using the same rules as slug-utils.mjs:
+// camelCase split, UPPERCASE-run split, letter↔digit split, underscore→hyphen,
+// lowercase, collapse hyphens, apply known overrides.
+function normalizeSegment(s) {
+  s = s.replace(/([a-z])([A-Z])/g, '$1-$2');
+  s = s.replace(/([A-Z]+)([A-Z][a-z])/g, '$1-$2');
+  s = s.replace(/([a-zA-Z])(\d)/g, '$1-$2');
+  s = s.replace(/(\d)([a-zA-Z])/g, '$1-$2');
+  s = s.replace(/_/g, '-');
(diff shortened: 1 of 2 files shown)


dannon:
Closing this since I think we're making it all work fine without it.

PR #3802, opened 2026-03-06
Pre-bake all redirects as S3 object metadata for native 301s

dannon:
## Summary

Astro in static mode generates `<meta http-equiv="refresh">` HTML pages for redirects. This means every redirect requires a full page load before the browser navigates to the canonical URL.

This PR pre-bakes all known redirects (from slug normalization, `redirects.yaml`, and frontmatter) as S3 object metadata (`x-amz-website-redirect-location`). The S3 website endpoint returns native 301 responses for these objects — no HTML page is loaded, no meta-refresh is parsed.

## Changes

- **`deploy/set-s3-redirects.mjs`**: New script that reads `generated-redirects.json` (built during the Astro build) and calls `aws s3api put-object` with `--website-redirect-location` for each redirect. Runs 20 concurrent requests. The key for each redirect is `{fromPath}/index.html` — matching where Astro puts the meta-refresh HTML.

- **`.github/workflows/publish.yml`**: Added a step after `aws s3 sync` that runs `set-s3-redirects.mjs`. Requires the existing `S3_BUCKET` env var to be set.

## Notes

- `generated-redirects.json` is produced by the build pipeline and contains ~1040 entries from all redirect sources.
- The script overwrites existing redirect objects on each deploy, so stale entries from removed redirects will persist until manually cleaned up (acceptable tradeoff).
- Pairs with a separate PR that adds real-time slug normalization to the CloudFront function for any URLs not covered by the pre-baked list.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,5 +30,7 @@ jobs:
           aws-region: us-east-2
       - name: Sync to preview S3
         run: aws s3 sync astro/dist/ s3://astro-hub/ --delete
+      - name: Set S3 redirect metadata
+        run: S3_BUCKET=astro-hub node deploy/set-s3-redirects.mjs
       - name: Invalidate CloudFront cache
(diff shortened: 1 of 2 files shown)


bgruening (approved):
Fancy

PR #3815, opened 2026-03-07
Normalize content directory names

dannon:
## Summary

Split out from #3804 — this is the content directory normalization work, separated from the subsite insert fix.

- Drop letter↔digit split rules from slug normalization and rename camelCase/underscore content directories to kebab-case
- Add CI lint check to catch non-normalized content directory names going forward

### DIFF
--- a/astro/package.json
+++ b/astro/package.json
@@ -80,5 +80,6 @@
     "prettier-plugin-astro": "^0.14.0",
     "tw-animate-css": "^1.4.0",
     "vitest": "^4.0.15"
-  }
+  },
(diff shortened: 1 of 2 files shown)


dannon:
## Pre-normalization sitemap snapshot

Built from `main` at 3954094e2 (pre-merge): https://gist.github.com/dannon/0bfc270430b48ba313e7083bf0a9573c

5396 URLs total.

## Sitemap comparison (main → this branch)

After rebasing onto main and rebuilding, the post-normalization build also produces 5396 URLs. 780 URL paths changed between the two builds:

- **~130 from actual directory renames** (camelCase/underscore → kebab-case) — these are covered by redirects in `redirects.yaml` and generated redirect HTML pages in `public/`
- **~650 from dropping the letter↔digit split rule** in the slug normalizer (e.g., `gcc-2012` → `gcc2012`) — these never existed on a deployed site, so no redirects needed

Every directory rename that was deployed has a corresponding redirect entry.

PR #3839, opened 2026-03-12
add some explanation to special handling of subsites with "global" and "all"

bgruening:
Trying to document the `global` and `all` subsites a bit.

Is that correct @dannon?

ping @natalie-wa @scottcain 

I think this is how its currently working. Do we want to keep this?

Do we have a better name as `global`? It has confused users in the past.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,9 @@ Events have additional fields:
 | `continent` | Two-letter code: AF, AS, AU, EU, GL, NA, SA          |
 | `location`  | Object with `name:` and optional `url:`              |
 | `contact`   | Event contact information                            |
-| `subsites`  | Array of subsites to display on, e.g. `[global, eu]` |
+| `subsites`  | Array of subsites to display on, e.g. `[all]` or `[eu]` |
+
+`all` and `global` are different. Use `all` to show a news item or event on every subsite feed. Use `global` for the root/global site only. A post tagged only with `global` will not appear on `/eu/`, `/fr/`, or other regional subsite feeds.
 
 ### Markdown
 


dannon:
So I dug into the actual filtering code to verify this, and the situation is messier than expected -- the codebase itself isn't consistent about what `global` means.

The short version: the shared content helpers in `content.ts` don't treat `global` as anything special when filtering for a subsite (they only check for the specific subsite name or `all`). But the bare pages and atom feeds have their own inline filtering that *does* include `global`-tagged content. So content tagged `[global, us]` shows up on EU feeds and bare pages but not on the main EU site pages. There's also a missing `all-eu` check in `eu/feed.atom.ts` that the other feeds have.

Before we can document the behavior, I think we need to decide what it *should* be. The most common pattern in content is `[global, us]` (~1k files) -- should that content appear on regional subsite pages or not? And if so, what's the actual difference between `global` and `all`?

I'd like to get this standardized into one shared filtering function and then document whatever we land on. What do you think?

bgruening:
What about:

`global` --> on the main site, but not on any other subsites, if not specified otherwise
`all` --> all sites, means global + all subsites
`all-eu` --> all EU subsites

If we follow this idea, `global` should maybe be renamed to `root` or `main` and I would expect that all news/events want to be listed on the main site. Exceptions might be news in French languages etc ...

natalie-wa:
> What about:
> 
> `global` --> on the main site, but not on any other subsites, if not specified otherwise `all` --> all sites, means global + all subsites `all-eu` --> all EU subsites
> 
> If we follow this idea, `global` should maybe be renamed to `root` or `main` and I would expect that all news/events want to be listed on the main site. Exceptions might be news in French languages etc ...

@bgruening @dannon 
I think this sounds like a good plan to me! Makes sense.

PR #3857, opened 2026-03-18
Fair update: add acknowledgement

Marie59:
Hi!

I tested it locally and the logos seem to spread a bit out of the pages and I don't get why.
Is it doing the same for you ?


Marie59:
hi @bgruening Is this good for you ?

PR #3858, opened 2026-03-18
Scottcain add gta25 banner

scottcain:


scottcain:
Note that I already have queued up a change that will have both the GCC and GTA logos on the page at the same time but I want this PR to merge first

PR #3912, opened 2026-04-09
Fix GCC page broken images, add GCC2022, fix logo layout

dannon:
## Summary

- Fix 4 broken image paths on `/gcc/` -- directory names had errant hyphens (`gcc-2019` vs `gcc2019`) that didn't match the actual image directories
- Add missing GCC2022 (Minneapolis) logo to the gallery
- Add CSS for `.inline-div` class (Gridsome leftover with no Astro definition) so logos render as a flexbox grid with consistent height instead of stacking vertically

## Test plan

- [ ] Verify all 16 conference logos render on `/gcc/`
- [ ] Confirm logos display in a wrapped grid layout with consistent height
- [ ] Check GCC2022 logo appears between GCC2023 and GCC2021

### DIFF
--- a/astro/src/styles/global.css
+++ b/astro/src/styles/global.css
@@ -253,6 +253,23 @@ body {
   margin-right: auto;
 }
 
+/* Inline image galleries (e.g. GCC logo grid) */
+.inline-div {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


bgruening:
Thanks @dannon

PR #3913, opened 2026-04-10
fix: use contributions field for GTN imports

arash77:
Replace the invalid `authors` field (not in schema-news.yaml) with `contributions.authorship` for news and `contributions.organisers` for events.

Adds a `CONTRIBUTORS.yaml` reverse-lookup to resolve RSS display names to contributor IDs. Falls back gracefully (no contributions block) when names don't match.

Fixes CI failures caused by the unknown `authors` field in imported frontmatter.

### DIFF
--- a/utils/gtn-import.py
+++ b/utils/gtn-import.py
@@ -23,35 +23,6 @@ with open(_SLUG_OVERRIDES_PATH) as f:
     _SLUG_OVERRIDES = json.load(f)
 
 
-_NEWS_CONTRIBUTION_KEYS = {"authorship", "funding"}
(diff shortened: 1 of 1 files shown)


bgruening:
I agree, this looks more like a hack, lets discuss it next week.

arash77:
This (https://github.com/galaxyproject/training-material/pull/6780) could help, which adds `contributions:<role>:<id>` categories directly to the GTN RSS feeds, so the contributor info is available on the feed side as well.

bgruening (approved):

bgruening:
Thanks @arash77

PR #3916, opened 2026-04-13
Fix events page links pointing to wrong detail pages

### LINKED ISSUE #3915: links on Events page are off by one
on https://galaxyproject.org/events/:

NGS Symposium 2026 points to https://galaxyproject.org/events/2026-04-08-eubi-datadays/

Small Scale Galaxy Admins Meeting ponts to https://galaxyproject.org/events/2026-03-31-dkfz-7th-ngs-symposium/

European Geosciences Union (EGU) 2026 points to https://galaxyproject.org/events/2026-04-small-scale/

dannon:
## Summary

Same Vue hydration mismatch that was fixed for the platforms page in #3821 -- SSR and client-side rendering produce different orderings, scrambling which href goes with which event card. Switches `EventsList` from `client:load` to `client:only="vue"` on all three events pages to bypass SSR entirely.

The tradeoff is that the event list renders client-side only instead of being in the initial HTML, but since it's an interactive/filterable component that requires JS anyway the practical difference is minimal. We can still follow up to identify the root cause of the hydration mismatch and restore SSR if we want, but this is a safe fix for now.

Fixes #3915

## Test plan

- [x] `npm run lint && npm run format:check` passes
- [x] `npm run build` succeeds
- [x] Playwright tests pass (139/141, 1 pre-existing failure unrelated to this change)
- [ ] Open `/events/` and verify event links point to correct detail pages
- [ ] Check subsite events pages (`/eu/events/`, `/freiburg/events/`)

### DIFF
--- a/astro/src/pages/[subsite]/events/archive/index.astro
+++ b/astro/src/pages/[subsite]/events/archive/index.astro
@@ -46,7 +46,7 @@ const eventData = events.map((e) => ({
 
     <div class="px-6 sm:px-8 py-6">
       <EventsList
-        client:load
+        client:only="vue"
         events={eventData}
         initialSubsite={subsiteId}
         showUpcoming={false}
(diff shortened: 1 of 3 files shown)


bgruening (approved):

bgruening:
Merci!

dannon:
Thank you!

hrhotz:
Thank You !

PR #3924, opened 2026-04-15
Add hot-reload watcher to make dev

arash77:
## What

Runs a content watcher in parallel with `astro dev` so edits to `content/` are preprocessed and picked up by Astro's HMR without restarting the dev server.

## Changes

**`npm run dev`** now spawns `preprocess.mjs --watch-only` alongside `astro dev`.

**`watchContent()`** is a complete content watcher that handles:

- **Edits** to existing pages/datasets — re-processes and writes the updated file to `src/content/`
- **New files** — detected immediately on save (previously silently ignored)
- **Deletions** — removes the corresponding generated file from `src/content/` to avoid phantom pages
- **Insert edits** — propagates to all parent pages that inline the insert via `<slot>`, using a reverse index built at startup
- **Ignored paths** — `CONTENT_IGNORE` constant shared between `preprocessContent` and `watchContent` so ignore patterns (`0examples/`, `navbar.yml`, `use/**/*.yml`, etc.) stay in sync

**Reliability fixes** so the watcher doesn't silently stop working:

- `fs.watch` + a **10-second fallback poll** — dropped inotify events (common on Linux with large directory trees) don't stall the watcher
- `fs.watch` failure is **non-fatal** — logs a warning and continues in poll-only mode
- **Concurrent scan guard** (`scanning`/`pendingScan` flags) — prevents two scans from racing to write the same files when events arrive faster than a scan finishes

## Test plan

```bash
make dev
```

- Edit an existing `content/.../index.md` → change appears in browser within ~200ms
- Create a new `content/news/2026/test/index.md` → page is available without restart
- Delete it → no stale file left in `src/content/news/`
- Edit an insert file (e.g. `content/admin/faq/questions.md`) → parent page updates
- Touch `content/0examples/vue-remark/index.md` → no `[watch] Processing` log line


bgruening:
@dannon can you please have a look at that one?

dannon:
### Code review

Nice -- the mtime-snapshot-on-startup is clean, the scan guard handles bursts fine, and returning both `.md`/`.mdx` candidates from `destPathsForMarkdown` to delete whichever exists is a tidy way around the components-toggle sibling problem. Sharing `CONTENT_IGNORE` between preprocess and watch is the right call too.

A few things before merge:

The P1 tests pick the first dir under the real `content/news/2025/` and plant files at the actual generated output paths ([test#L407-L420](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.watch.test.mjs#L407-L420), [L468-L490](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.watch.test.mjs#L468-L490)) -- if a test crashes mid-run you've got corrupted output until the next full preprocess. Same story with P2 ([L500-L525](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.watch.test.mjs#L500-L525)) writing to `content/test-inserts/` and `content/test-insert-parent/` in the live tree. There's a `tmpDir` allocated that never gets used -- could just do the whole thing in there. Also `newsFiles[0].name` will blow up if 2025 is empty.

`SLOT_RE` is defined twice ([preprocess.mjs#L874](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.mjs#L874), [L914](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.mjs#L914)) -- pull it up to module scope.

PR body still says \"3-second fallback poll\" but [the code is 10s](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.mjs#L1061) -- looks like the bump commit didn't update the description.

The reverse-index keys on the raw path ([preprocess.mjs#L962](https://github.com/galaxyproject/galaxy-hub/blob/d3abc5158d3b77906ac212677db5e9fd7f6213a9/astro/src/build/preprocess.mjs#L962)), so if anyone ever uses the denormalized slot fallback that `resolveInsertContent` supports (`<slot name=\"/ifb/lead-1\" />` -> `lead1.md`), insert hot-reload would silently miss. Nothing in content uses it today, just worth a note so the next person doesn't get bitten.

Minor: the `(trap 'kill 0' EXIT; ... & astro dev)` is POSIX-only, won't work on Windows. Probably fine but `concurrently` is the portable option if anyone cares.

Mostly the test isolation is what I'd want fixed -- everything else is small.

arash77:
Thanks for the thorough review @dannon! Addressed everything in 352acb5:

**Test isolation (P1 & P2)** — The main fix: `processMarkdownFile`, `destPathsForMarkdown`, `getContentCollection`, `inlineInserts`, and `resolveInsertContent` now all accept an optional `contentDir` (and `outputDir` for the two path functions), defaulting to the module-level constants so existing call sites are unaffected. P2 tests now build their mini content tree entirely inside `os.tmpdir()` and pass `{ contentDir: tmpDir, outputDir: tmpDir }` — no live-tree writes at all, and `afterEach` is a single `rmSync(tmpDir, { recursive: true })` which is as crash-safe as it gets. P1 got an `afterEach` that always removes the generated output files, plus a `newsFiles.length` guard so an empty `content/news/2025/` won't blow up.

**`SLOT_RE` duplicate** — Hoisted to module scope; both use sites now reset `lastIndex` before the `while` loop.

**Reverse-index comment** — Added a note at the key-building lines flagging the denormalized slot alias blind spot.

**PR description** — Updated "3-second fallback poll" → "10-second".

The `(trap 'kill 0' EXIT; ...)` / `concurrently` note is fair — left as-is for now since the project doesn't otherwise depend on Windows dev support, but happy to swap it if needed.

dannon:
Thanks for the updates, this looks good to me now.  I pushed a minor tweak to fully isolate P1.

dannon (approved):

PR #3925, opened 2026-04-15
Adding a fellowship awardee's photo

natalie-wa:
Uploading Shubham's headshot for adding to the GCC2026 website.


bgruening:
@natalie-wa do we need a second image here?

natalie-wa:
> @natalie-wa do we need a second image here?

@bgruening A second? I don't think so.

PR #3930, opened 2026-04-17
Update input fields to textareas in instance request template

ccoulombe:
Changed input fields to textareas for better user input.

Changed
- Public code repositories (org or list)
- CVMFS reference data availability

items so we can provide lists or a bit more information as needed.


bgruening (approved):

bgruening:
Thanks!

PR #3931, opened 2026-04-18
feat(eu): port About page from usegalaxy-eu/website

itisAliRH:
## Summary

- Adds `content/eu/about/index.md` ported from [usegalaxy-eu/website about.md](https://github.com/usegalaxy-eu/website/blob/master/about.md)
- Converts Font Awesome icons to Lucide (`<Icon name="..." />`)
- Co-locates `tools.png` screenshot alongside the page
- Updates `content/eu/navbar.yml` to point People/Tools/About links to hub-relative URLs instead of old usegalaxy-eu.github.io

## Test plan

- [ ] `make content-lint && make validate-metadata` pass
- [ ] Dev server shows `/eu/about/` with proper content and tool screenshot
- [ ] Navbar "About → The European Server" link goes to `/eu/about/`

Part of [usegalaxy-eu/issues#943](https://github.com/usegalaxy-eu/issues/issues/943)


content/eu/navbar.yml
bgruening, line 1: A few changes in here, should probably not be here.

bgruening:
@itisAliRH I tried to fix the navbar loading - which was broken on the Hub for subdomains. I then also fixed the links in the navbar which is now loaded from the navbar.yml file.

I also fixed the build error in your PR.

bgruening:
@dannon since this is messing with the navbar, do you want to take a look?

dannon:
Nice work getting the navbar YAML-driven so subsites don't need Vue edits to customize. A couple of things while I was poking around:

The new `detectSubsiteFromPath` in `astro/src/utils/navbar.ts` looks identical to the one already exported from `astro/src/stores/subsiteStore.ts` -- can we just import it from the store instead of keeping two copies? Worried they'll drift when someone adds a subsite and only updates one.

Also noticed `content/eu/navbar.yml` mixes conventions now -- News/Events have `relativeTo: subsite` with `target: eu/news/`, while People drops `relativeTo` and uses `/eu/people/`. Both end up at the same URL because `prefixSubsiteHref` skips re-prefixing when the prefix is already there, but is there a reason to keep `relativeTo` on the ones that already include `eu/`? Feels like we should pick one style so the next person porting a subsite navbar isn't guessing.

Minor stuff on the About page itself -- a bunch of H1s (`# usegalaxy.eu`, `# Data Analysis...`, etc.) plus the frontmatter title. Want to demote those to `##` while we're in here, or keep it a faithful port for now? And the 2-space indent on the ACKNOWLEDGEMENT paragraphs is a Jekyll holdover, probably safe to strip.

Not really blocking on my end, but it'd be nice to polish those few things up.

itisAliRH:
@bgruening Thanks for fixing the broken parts.

@dannon Thank you for the comments. All mentioned items are fixed now.
I think using the prefixed URLs is more readable, so I dropped the `relativeTo`.

bgruening (approved):

PR #3932, opened 2026-04-18
feat(eu): add Tools page + weekly CI from usegalaxy-eu/website

itisAliRH:
## Summary

- Adds `content/eu/tools/index.md` with initial snapshot of 4219 tools from usegalaxy.eu

|Before|After|
|---|---|
|image|image|

- Ports `update_tools.py` as `scripts/update-tools.py` (output target changed; handles `Tool`/`ToolSection`/`ToolSectionLabel` model classes).
- Adds `.github/workflows/update-eu-tools.yml` (cron Sunday 23:00; uses `peter-evans/create-pull-request@v7`)
- Includes global styling for tool lists in `astro/src/styles/global.css`.
- Updates `content/eu/navbar.yml` to hub-relative URLs

## Test plan

- [ ] `make content-lint && make validate-metadata` pass
- [ ] Dev server shows `/eu/tools/` with tool buttons
- [ ] After merge: `gh workflow run update-eu-tools.yml` completes and opens a PR

Part of [usegalaxy-eu/issues#943](https://github.com/usegalaxy-eu/issues/issues/943)


content/eu/navbar.yml
bgruening, line 1: Those changes should not be in here I guess. But also this file does not seem to have any effect. Probably a HUB bug.

bgruening (commented):
From the style, https://usegalaxy-eu.github.io/tools looks better than this version because the tools are separated from each other with those boxes.

scripts/update-tools.py
bgruening, line 1: maybe rewrite this script slightly, so that it can take the server as parameter ... with the aim that if other domains/servers wants to have a similar side, they could simply add this to the CI with the same script

bgruening (approved):

PR #3934, opened 2026-04-18
feat(eu/bare): port subdomain front pages – batch 1

itisAliRH:
## Summary

Ports bare iframe landing pages from `usegalaxy-eu/website` for: **cancer, humancellatlas, rna, singlecell, clipseq, spatialomics, eirene**.

Per-page migration work:
- Adds the shared bare-page notice slot: `<slot name="/bare/eu/usegalaxy/notices" />`
- Removes Jekyll-only syntax and old MDX-breaking comments
- Converts old tool-template links to static links where mappings still resolve
- Replaces stale/broken template output with plain text where no valid tool mapping exists
- Vendors migrated media under `content/assets/media/usegalaxy/...`
- Rewrites old `*.usegalaxy.eu/assets/media/...` image URLs to Hub-owned asset paths
- Keeps the pages lightweight for iframe use

Notes:
- `components: true` was removed where it was not needed by the page.
- A small Playwright timing fix for the citations year filter is included because it was flaky during full-suite runs.

## Test plan

- [x] `make content-lint`
- [x] `make validate-metadata`
- [x] `npm run lint`
- [x] `make build`
- [x] `npm run test:unit`
- [x] `npm run test`
- [x] Local spot-check of representative `/bare/eu/usegalaxy/<topic>/` pages

Part of https://github.com/usegalaxy-eu/issues/issues/943


bgruening:
`make build` is failing for me locally, because of wrong "comments" in the mdx files. I tried to fix that in the last commit.

The RNA subpage has a lot of magic in, which we would need to convert differently. There is a template in it which creates tool links. Please talk to @pavanvidem if this is still needed. You will see this pattern in other pages I think.

Furthermore, there are images that are not resolving:

e.g. 
``` 

![single cell Galaxy](https://singlecell.usegalaxy.eu/assets/media/logo_single_cell.svg)
![Human Cell Atlas](https://singlecell.usegalaxy.eu/assets/media/hca.png)
``` 

the images should be moved to this github repo.

content/bare/eu/usegalaxy/spatialomics/index.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Galaxy Spatial Omics
```
bgruening, line 2: do we need "components: true"

itisAliRH: > do we need "components: true"

no. that was leftover from the bulk conversion/normalization pass. The page only needs the bare notices slot and static HTML/Markdown content, so it builds without enabling component rendering.

itisAliRH:
- The MDX comment issues are fixed; there are no raw HTML/Jekyll comments left in these batch 1 bare pages.
- The old RNA tool-link template/include logic has been replaced with static links, where the old tool mapping still resolved. For a few old scater entries that no longer had a usable mapping, I left plain text rather than creating broken links.
- The missing images are now vendored into this repo under `content/assets/media/usegalaxy/...`, including the single-cell logo and HCA images, and the page references were updated away from `*.usegalaxy.eu/assets/media/...`.

content/assets/media/usegalaxy/humancellatlas/hca.png
bgruening, line 1: This image is now twice in this PR, should we reuse the other one?

content/bare/eu/usegalaxy/cancer/index.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Galaxy Cancer
+components: true
```
bgruening, line 3: why is that here needed? because of the one comment?

I guess it does not harm, just interested.

bgruening (approved):

content/bare/eu/usegalaxy/cancer/index.md
itisAliRH: it’s needed because this page uses `<Icon name="graduation-cap" />`. The JSX comment only matters once the page is MDX; it is not the main reason for the flag.

PR #3943, opened 2026-04-22
Remove auto-assignment via template and code ownership

jdavcs:
As per recent discussion on slack.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,3 @@
 content/community/coc/      @assuntad23 @beatrizserrano @frederikcoppens @hexylena @fubar2
 content/news/               @natwhitaker
 content/events/             @natwhitaker
-**/designations.csv         @mschatz @bgruening @bebatut @nekrut @GarethPrice-Aus @blankenberg @hujambo-dunia @nakucher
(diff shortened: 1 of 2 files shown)


bgruening:
What was the discussion on Slack?

jdavcs:
> What was the discussion on Slack?

Here's the link: https://app.slack.com/client/T0262EGNV/C02HJDML6

The consensus was that auto-assignming multiple people (including all the PIs in this particular case) usually doesn't work as expected. For the submitter of the issue or PR the message that they "...assigned these people to the issue/pr" may be suprising if not awkward. Also, there's a high likelyhood of the issue nor being addressed in a timely manner (too many assignees, etc.). Essentially, if one really wants someone assigned, doing that manually is a better option.

bgruening:
Ok, so we don't review those PRs now? Or we trust that the 2-3 people to review them that usually merge?

jdavcs:
> Ok, so we don't review those PRs now? Or we trust that the 2-3 people to review them that usually merge?

We treat them the same as other PRs, I suppose? As far as I remember, we usually just merge updates to the hub. If it's a sensitive change (for example, a change to policy reflected in the dot-star application form), we can manually assign a person or ask on the channel. Other than that, same as other PRs.

bgruening (approved):
Ok!

PR #3952, opened 2026-04-29
Add blogpost for workflows using TabPFN newer models

anuprulez:
Workflows for:
- Splice junction prediction
- Diabetes detection


SaimMomin12 (approved):
Thanks @anuprulez

PR #3976, opened 2026-05-13
Update index.md to try fixing display issue + add content

yvanlebras:
- Fix div section issues trying to make differently to avoid having display issue on Galaxy hub
- Add some content so we can also argue that Galaxy can be used to facilitate accessing and generating data and metadata with high FAIRness degree
- Add me as co-author if relevant


Marie59 (approved):
LGTM !
Thanks for the fix !

PR #3996, opened 2026-05-21
ci: stop image compression bot from self-triggering PR loops

itisAliRH:
## Summary

Fixes the image compression workflow loop seen on #3937, where `galaxy-hub-bot` repeatedly committed optimized images and posted a new Calibre comment for each follow-up run.

The root cause is that this workflow intentionally uses a GitHub App installation token. Per GitHub Actions docs, workflow changes made with `GITHUB_TOKEN` do not usually create new workflow runs, but changes made with a GitHub App token or PAT do. That means every Calibre commit on a PR can trigger another `pull_request.synchronize` run of the same compression workflow.

This PR:
- skips the compression job for PR events authored by `galaxy-hub-bot` / `galaxy-hub-bot[bot]`
- updates `calibreapp/image-actions` from `1.4.1` to `1.5.0`
- removes the custom `minPctChange: '2.5'` override so Calibre’s safer defaults apply
- adds workflow concurrency to avoid overlapping compression runs on the same PR/ref
- adds WebP/AVIF path filters to match the image types supported by Calibre

## References

- Affected PR: #3937
- GitHub docs: triggering workflows from workflows with `GITHUB_TOKEN` vs GitHub App/PAT tokens  
  https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow
- GitHub docs: workflow concurrency and `cancel-in-progress`  
  https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency
- Calibre image-actions README  
  https://github.com/calibreapp/image-actions
- Calibre image-actions `1.5.0` release notes, adding `minAbsChange` for tiny repeated lossy PNG reductions  
  https://github.com/calibreapp/image-actions/releases/tag/1.5.0

## Local verification

- Parsed workflow YAML locally.
- Ran `git diff --check`.
- Simulated GitHub PR payloads locally:
  - normal user-authored `pull_request.synchronize` remains allowed
  - bot-authored `pull_request.synchronize` is blocked
  - `push`/scheduled/manual flows remain allowed
- Ran `act` against a bot-authored PR event and confirmed the job is skipped before any workflow steps execute.

### DIFF
--- a/.github/workflows/optimize-images.yml
+++ b/.github/workflows/optimize-images.yml
@@ -43,13 +43,12 @@ jobs:
     if: |
       github.repository == 'galaxyproject/galaxy-hub' &&
       (github.event_name != 'pull_request' ||
-       (github.event.pull_request.head.repo.full_name == github.repository &&
-        github.actor != 'galaxy-hub-bot' &&
-        github.actor != 'galaxy-hub-bot[bot]'))
+       github.event.pull_request.head.repo.full_name == github.repository)
     steps:
       - name: Checkout Branch
         uses: actions/checkout@v6
       - name: Generate GitHub App token
+        if: github.event_name != 'pull_request'
         id: app-token
         uses: actions/create-github-app-token@v3
         with:
@@ -59,7 +58,9 @@ jobs:
         id: calibre
         uses: calibreapp/image-actions@1.5.0
         with:
-          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          # PR runs use GITHUB_TOKEN to avoid recursively triggering this workflow.
+          # Non-PR runs use the App token so create-pull-request can publish updates.
+          GITHUB_TOKEN: ${{ github.event_name == 'pull_request' && github.token || steps.app-token.outputs.token }}
           webpQuality: '90'
           # Non-PR events need compress-only mode so create-pull-request can publish the result.
           compressOnly: ${{ github.event_name != 'pull_request' }}
--- a/.github/workflows/optimize-images.yml
+++ b/.github/workflows/optimize-images.yml
@@ -39,28 +39,35 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
-    # Only run on main repo on and PRs that match the main repo.
+    # Only run on main repo and on PRs that match the main repo.
+    # Skip when the bot's own compression commit re-triggers this workflow
+    # (synchronize + bot actor) to break the infinite loop. Safe for
+    # bot-authored PRs (e.g. GTN imports) because those trigger as
+    # 'opened', not 'synchronize'.
     if: |
       github.repository == 'galaxyproject/galaxy-hub' &&
       (github.event_name != 'pull_request' ||
-       github.event.pull_request.head.repo.full_name == github.repository)
+       github.event.pull_request.head.repo.full_name == github.repository) &&
+      !(github.event_name == 'pull_request' && github.event.action == 'synchronize' && startsWith(github.actor, 'galaxy-hub-bot'))
     steps:
       - name: Checkout Branch
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - name: Generate GitHub App token
-        if: github.event_name != 'pull_request'
         id: app-token
         uses: actions/create-github-app-token@v3
         with:
           client-id: ${{ secrets.HUB_BOT_APP_CLIENT_ID }}
           private-key: ${{ secrets.HUB_BOT_APP_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+          permission-metadata: read
       - name: Compress Images
         id: calibre
         uses: calibreapp/image-actions@1.5.0
         with:
-          # PR runs use GITHUB_TOKEN to avoid recursively triggering this workflow.
-          # Non-PR runs use the App token so create-pull-request can publish updates.
-          GITHUB_TOKEN: ${{ github.event_name == 'pull_request' && github.token || steps.app-token.outputs.token }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
           webpQuality: '90'
           # Non-PR events need compress-only mode so create-pull-request can publish the result.
           compressOnly: ${{ github.event_name != 'pull_request' }}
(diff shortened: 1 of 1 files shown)


bgruening (approved):
Some tests are failing.

itisAliRH:
> Some tests are failing.

@bgruening The failure is not related to these changes. The Playwright artifact screenshots show that both pages had actually rendered content by the time the timeout happened.

arash77 (commented):
Instead of skipping the entire workflow for any action by the bot (which would also prevent image optimization on PRs created by the bot for other purposes, such as in PR #3997), we can check the commit message. If the last commit message contains "optimised images" or "compress images" (case-insensitively), we skip the core optimization steps.

This review contains two inline suggestions:
1. Revert the global `github.actor` filter from the job-level `if:` block.
2. Update the checkout step to use `fetch-depth: 2` (so we can inspect the merge commit history) and add a `Check for loop` step.

After applying these suggestions, you will also need to add the following conditional to the remaining steps so they only run if the loop check did not skip:

`if: steps.check-loop.outputs.skip != 'true'`

Here is how those steps should look:

```yaml
      - name: Generate GitHub App token
        if: steps.check-loop.outputs.skip != 'true'
        id: app-token
        uses: actions/create-github-app-token@v3
        ...

      - name: Compress Images
        if: steps.check-loop.outputs.skip != 'true'
        id: calibre
        uses: calibreapp/image-actions@1.5.0
        ...

      - name: Create Pull Request
        if: |
          steps.check-loop.outputs.skip != 'true' &&
          github.event_name != 'pull_request' &&
          steps.calibre.outputs.markdown != ''
        uses: peter-evans/create-pull-request@v8
        ...
```

itisAliRH:
Thank you @arash77!

I changed this to the cleaner approach and pushed the changes: PR image-compression runs now use the default `GITHUB_TOKEN`, while non-PR runs still generate and use the GitHub App token for `create-pull-request`. That avoids the recursive `pull_request.synchronize` loop using GitHub’s built-in `GITHUB_TOKEN` behavior, without blocking unrelated bot-authored PR updates.

Does that make sense?🤔

arash77 (commented):
## Simpler approach: job-level `if` + App token for all runs

The current approach (using `GITHUB_TOKEN` for PR runs) prevents the loop via GitHub's built-in behavior, but it has a side effect: commits pushed with `GITHUB_TOKEN` don't trigger other workflows. Since `main.yml` (Lint, Unit, Playwright) is a required status check, non-admin contributors will be blocked from merging — the checks stay in "Expected / Pending" indefinitely because they never run on the bot's commit.

**The fix**: Keep the App token for all runs (so status checks always trigger) and instead skip the entire job when the bot's own commit re-triggers it. This is cleaner than checking commit messages because:

1. No `fetch-depth: 2`, no shell scripting, no step outputs
2. The job doesn't even start — no checkout, no token generation, nothing
3. Safe for bot-authored PRs (e.g. GTN imports) because those trigger as `opened`, not `synchronize`

The three suggestions below apply this approach. See the inline comments for details.

.github/workflows/optimize-images.yml
```diff
@@ -38,19 +48,20 @@ jobs:
       - name: Checkout Branch
         uses: actions/checkout@v6
       - name: Generate GitHub App token
+        if: github.event_name != 'pull_request'
```
arash77:
```suggestion
```

Remove this `if` — the App token should be generated for **all** runs (including PRs) so that `main.yml` status checks are triggered on the bot's commits. The loop prevention moves to the job-level `if` instead (see review body).

.github/workflows/optimize-images.yml
```diff
@@ -38,19 +48,20 @@ jobs:
...
       - name: Compress Images
         id: calibre
-        uses: calibreapp/image-actions@1.4.1
+        uses: calibreapp/image-actions@1.5.0
         with:
-          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
-          # Attempt to prevent webp recompression by minimizing diff.
+          # PR runs use GITHUB_TOKEN to avoid recursively triggering this workflow.
+          # Non-PR runs use the App token so create-pull-request can publish updates.
+          GITHUB_TOKEN: ${{ github.event_name == 'pull_request' && github.token || steps.app-token.outputs.token }}
```
arash77:
```suggestion
          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```

Use the App token for all runs. With the loop prevention moved to the job-level `if` (see review body), we no longer need to switch to `GITHUB_TOKEN` for PR runs. Using the App token ensures `main.yml` status checks are triggered on every commit, so non-admin contributors aren't blocked by branch protection.

arash77 (commented):
## Simpler approach: job-level `if` + App token for all runs

The current approach (using `GITHUB_TOKEN` for PR runs) prevents the loop via GitHub's built-in behavior, but it has a side effect: commits pushed with `GITHUB_TOKEN` don't trigger other workflows. Since `main.yml` (Lint, Unit, Playwright) is a required status check, non-admin contributors will be blocked from merging — the checks stay in "Expected / Pending" indefinitely because they never run on the bot's commit.

**The fix**: Keep the App token for all runs (so status checks always trigger) and instead skip the entire job when the bot's own commit re-triggers it. Add this to the job-level `if:`:

```yaml
    # Only run on main repo on and PRs that match the main repo.
    # Skip when the bot's own image commit re-triggers this workflow (synchronize + bot actor),
    # which breaks the infinite loop. Safe for bot-authored PRs (e.g. GTN imports)
    # because those trigger as 'opened', not 'synchronize'.
    if: |
      github.repository == 'galaxyproject/galaxy-hub' &&
      (github.event_name != 'pull_request' ||
       github.event.pull_request.head.repo.full_name == github.repository) &&
      !(github.event_name == 'pull_request' && github.event.action == 'synchronize' && startsWith(github.actor, 'galaxy-hub-bot'))
```

**Why this is cleaner:**
- No `fetch-depth: 2`, no shell scripting, no step outputs, no `env` var
- The job doesn't even start on bot-retrigger — no checkout, no token generation, nothing
- Safe for bot-authored PRs (GTN imports) because those trigger as `opened`, not `synchronize`
- App token is used for all runs → `main.yml` always triggers → status checks pass → PRs remain mergeable

The two inline suggestions apply the step-level changes (remove the `if` on token generation, use App token unconditionally).

bgruening (approved):

bgruening:
Thanks, lets try.

PR #4007, opened 2026-05-28
Add a BOFs tab to the GCC26 page

Marie59:
Hi all, 
Sorry to bother With the new look of the community hub, I totally forgot how I can add this new tab to the GCC website. I created the content, I remember that for it to show I need to go sowhere in the astro part but I don't know where.

Anyone can help me ?

Thanks !!

### DIFF
--- a/content/events/gcc2026/bofs/index.md
+++ b/content/events/gcc2026/bofs/index.md
@@ -1,0 +1,15 @@
+# BoFs: Birds of Feathers  
+
+BoFs are a great opportunity to gather with other Galaxy community members around shared topics, ideas, questions, or challenges. 
+If there is something you would like to discuss with others during the conference, we encourage you to submit a BoF idea.
+
+These moments of discussions on a specific topic will be held Monday 22nd from 5 to 6pm and Wednesday 24th from 5:15 to 6:30pm.
+
+The Sessions can be structured in whatever way works best for the topic, whether that includes short presentations, a guided discussion, or an open conversation. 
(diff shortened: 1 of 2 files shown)


dannon:
Hey! I can help. You don't need to register the tab anywhere -- the GCC tabs are auto-generated from the subdirectories under `content/events/gcc2026/`, so just dropping in `bofs/index.md` is enough to make it show up.

I pushed a few small tweaks to your branch:
- Added `nav_title: BoFs` to the frontmatter so the tab reads "BoFs" -- without it the auto title-casing turns it into "Bofs". Also `autotoc: false` to match the other GCC pages.
- Your submit link was using image syntax (``), which renders as a broken image instead of a link -- swapped it for a proper button like the one on the Register page.
- Minor wording fix: "Birds of Feathers" -> "Birds of a Feather".

Built it locally and the BoFs tab now shows up between Abstracts and CoFest, with the button linking out to your form. Should be good to go!

Marie59:
Thank you @dannon !! 

There's a mistake regarding on of your change ` ["Key 'nav_title' was not defined. Path: '/gcc2026/bofs'"]` I don't know how to fix it

dannon:
That one's on me -- `nav_title` is a valid key the nav code uses to label the tab, but it was never added to the metadata validation schema, so it tripped the check. Just pushed a commit adding it to the schema, so Metadata Validation should go green once CI re-runs. Nothing to do on your end!

PR #4003, opened 2026-05-25
Add Release Guardians Page

guerler:
I picked up on some of the concepts we floated during a recent JHU team meeting, including the name: Release Guardians. Release Guardians is a new release-validation workflow for Galaxy that adds to the traditional email-based testing coordination model with a lightweight, public, GitHub-label-driven community process.

PRs included in a release cycle are tagged with `release-testing-<version>` (for example `release-testing-26.1`). Community members claim a PR by adding `release-testing-in-progress`, exercise the change on the Galaxy Test Server or any other deployment they have access to, open issues for discovered regressions or problems, and finally replace the label with `release-testing-complete` once validation is finished. There is no signup system, no separate dashboard, and no project-board layer. GitHub labels are the operational source of truth.

A public page at `/community/release-guardians/` is generated from live GitHub state using a small Node sync script and an Astro event template. PRs are partitioned into **Needs Validation**, **In Progress**, and **Complete** sections. Each card surfaces the PR title, labels, extracted summary, and — when applicable — the contributor currently acting as Guardian for that PR. The implementation is intentionally lightweight: one config file per release cycle, one template, one sync script, and one scheduled GitHub Actions workflow that regenerates the static page every 12 hours as a PR to be merged. No backend services, database, authentication layer, or signup flow are involved.

The goals are to make release testing more visible, scalable, collaborative, and community-driven while reducing the repeated coordination overhead of assembling temporary testing groups from scratch for every release cycle.

### First Cycle: Galaxy 26.1

The first Release Guardians cycle covers the Galaxy 26.1 release.

- **Testing window:** 2026-06-01 → 2026-06-05
- **In-scope PRs:** `release-testing-26.1` on `galaxyproject/galaxy`
- **Live page:** `/community/release-guardians/`
- **Archived event page:** `/events/2026-06-01-release-guardians-26-1/`
- **Primary test deployment:** Galaxy Test Server
- **Async coordination:** Release Testing Matrix room
- **Live coordination:** Testing and Hardening Working Group meeting
- **Organisers:** `@guerler`, `@mvdbeek`

Subsequent release cycles only require updating a single config file (`astro/src/data/release-guardians-config.json`) with the new version, labels, and dates. Previous cycles remain available as a permanent public archive documenting how each Galaxy release was validated.

cast

PR #4009, opened 2026-05-29
Allow contributors without label permissions to indicate testing state

guerler:
Follow up #4003. As mentioned during the discussion, one limitation of the labeling approach is that some contributors can participate in release testing but do not have permission to apply labels on Galaxy PRs. This change allows the canonical release testing phrases to be posted as PR comments as an alternative.


mvdbeek:
No, please don't make people spam the PRs with this.

guerler:
Fair enough, the intent was to make release testing participation possible for contributors without label permissions. We can keep labels as the canonical workflow and restrict participation to testers with permissions for now.

PR #4017, opened 2026-06-01
Add Orbit documentation page

nekrut:
## What

Adds an **Orbit** documentation page, the desktop co-scientist shell for the Loom agent.

- **`/orbit/`** — a permanent Hub page describing Orbit (install, LLM provider/model setup, Galaxy routing, notebook, slash commands, getting help).
- **Navigation** — an `Orbit` tab in the masthead (`content/navbar.yml`) and the default sidebar top links (`astro/src/utils/navbar.ts`).

## Notes

Supersedes nekrut/galaxy-hub#20, which was mistakenly merged into the fork rather than opened here. Branch is based cleanly on `galaxyproject/galaxy-hub:main` and contains only the doc additions (no unrelated fork drift, no announcement post).


astro/src/utils/navbar.ts
```diff
@@ -34,6 +34,7 @@ export interface SidebarNavigation {
 const DEFAULT_TOP_LINKS: SidebarLink[] = [
   { label: 'News', href: '/news/' },
   { label: 'Events', href: '/events/' },
+  { label: 'Orbit', href: '/orbit/' },
```
bgruening:
```suggestion
  { label: 'Orbit', href: '/tools/orbit/' },
```

Should that be sub-level entry, something like /tools/orbit/ or is tools in the Galaxy space not a good fit.
I could imaging we have many more such things coming up that we want to promote, similar to /projects/covid19/ etc...

nekrut: Good call — applied. Moved the page to `content/tools/orbit/` so it lives at `/tools/orbit/`, and updated both nav references (`navbar.yml` and `navbar.ts`). Agreed this gives us a clean namespace for future promoted entries.

nekrut:
**Note on the FR News test change**

This PR is documentation-only, but it also includes one small test fix in `astro/tests/bare-pages.spec.ts`. The `/bare/fr/news/` Playwright test was already failing on `main`: the page now renders two `<h2>` headings ("Galaxy France News" and "Featured integrations"), so the `page.locator('h2')` assertion hit a strict-mode violation (two matches). I scoped it to the heading by accessible name (`getByRole('heading', { name: 'Galaxy France News' })`) so it asserts the specific heading rather than all `h2`s. Verified locally: `1 passed`. Happy to split it into a separate PR if you'd prefer.

---

**Note à propos du changement sur le test « FR News »**

Cette PR ne concerne que la documentation, mais elle inclut aussi une petite correction de test dans `astro/tests/bare-pages.spec.ts`. Le test Playwright de `/bare/fr/news/` échouait déjà sur `main` : la page affiche désormais deux titres `<h2>` (« Galaxy France News » et « Featured integrations »), si bien que l'assertion `page.locator('h2')` provoquait une violation du mode strict (deux correspondances). Je l'ai ciblée sur le titre par son nom accessible (`getByRole('heading', { name: 'Galaxy France News' })`) afin de vérifier ce titre précis plutôt que l'ensemble des `h2`. Vérifié en local : `1 passed`. Je peux la déplacer dans une PR distincte si vous préférez.

dannon:
Really solid doc -- I checked most of the factual claims against the loom source and the v0.1.1 release, and nearly everything lines up exactly: the slash commands, keyboard shortcuts, the 900px files-panel breakpoint, the routing tags, the `loom.managed` git behavior, the skills allowlist + 24h cache, the WSL/safeStorage note, and the model IDs (including the ones that looked too new -- `o3`, `gpt-5.4`, the gemini-3.x entries -- they're all real in the registry Orbit resolves against). The `/tools/orbit/` namespacing and nav wiring look right too.

A few things I'd tweak:

**Stale version pin (should fix).** The Intel-Mac install commands hardcode `v0.1.0-alpha.5` (the clone and the tarball), but the current release is v0.1.1 -- so anyone copy-pasting those builds an old alpha. Everywhere else you use `<version>` placeholders; just bump these two blocks.

**New page leans on Bootstrap (worth considering).** The page uses `alert`, `table`/`table-responsive`, and `--bs-border-color`, which are the legacy Bootstrap 4 classes the `.bs-compat` shim exists to carry for the old content (DESIGN.md:399). It renders fine, but it gets marked `.bs-compat`, which adds a brand-new page to the legacy bucket we're trying to shrink. Might be worth using the modern callout/Tailwind pattern instead.

**"Pi:" in the resume example.** That looks like the underlying pi-ai framework name leaking through -- the rest of the doc calls it Orbit/the agent, and the UI doesn't label it "Pi". Probably want `Orbit:` there.

Couple of tiny nits: the RPM asset is actually `orbit-<version>-1.x86_64.rpm` (forge adds the `-1`), and under the Google provider the exact id is `gemini-3.1-pro-preview` (the bare `gemini-3.1-pro` still resolves via substring match, so no big deal).

I can fix these if you want me to make a pass.

nekrut:
Thanks for the careful review, @dannon — all addressed in f215c62:

- **Stale version pin** — Intel-Mac install (clone + tarball) bumped from `v0.1.0-alpha.5` to `v0.1.1`.
- **Bootstrap / `.bs-compat`** — replaced the `alert` divs and the `table`/`table-responsive` examples block with the modern Tailwind callout/card patterns (`border-l-4 border-l-galaxy-gold|primary`, `bg-white rounded-lg shadow-sm`, `divide-y`). Confirmed the preprocess no longer injects a `.bs-compat` marker, so the page stays out of the legacy bucket. The remaining tables are plain Markdown pipe-tables, which render classless and don't trigger the marker.
- **"Pi:" → "Orbit:"** in the resume example.
- **RPM asset name** → `orbit-<version>-1.x86_64.rpm` (the `-1` forge adds), in all three spots.
- **Google id** → exact `gemini-3.1-pro-preview`.

Nice catch on `o3`/`gpt-5.4`/gemini-3.x being real in the registry — those were intentional.

nekrut:
Correction to my earlier comment: I was working from a stale branch and didn't realize @dannon's points were **already resolved on `main`** after this PR merged — `6cfc6dedd` ("Fix a few accuracy nits") covers the version bump, `gemini-3.1-pro-preview`, the RPM `-1` suffix and the `Orbit:` label, and `0a7a9b6d0` ("De-Bootstrap the Orbit docs page") replaced the Bootstrap `alert`/`table` with the native `callout`/`gx-tile-grid` components (cleaner than the raw Tailwind I'd pushed). So those fixes are in, and better. My follow-up commit was redundant; I've deleted the branch. Thanks @dannon for the pass.

PR #4043, opened 2026-06-05
Source the brand palette from @galaxyproject/brand-tokens

dannon:
The hub's `@theme` block has been the de-facto canonical home of Galaxy's brand palette -- which is awkward, because other properties (the Keycloak login theme, for one) keep hand-mirrored copies that quietly drift. This extracts that palette into a standalone package, [`@galaxyproject/brand-tokens`](https://github.com/galaxyproject/galaxy-brand-tokens), and points the hub at it so there's one source of truth everyone consumes.

What changes:
- `global.css` drops the inline `@theme { … }` block in favor of `@import "@galaxyproject/brand-tokens/theme.css"`. The package ships the exact `@theme` source lifted from here, so the generated Tailwind utilities (`bg-bay-of-many-700`, etc.) are byte-for-byte unchanged.
- `DESIGN.md` slims to hub-specific guidance; the brand-level palette/typography tables move to the new repo's canonical `DESIGN.md`, linked from the top.
- Adds the dependency at `^0.1.0` plus the matching lockfile entry.

The package is published as `0.1.0`, and a CI drift-guard keeps its generated CSS in lockstep with the source. A companion PR migrates the Keycloak login theme to fetch the same tokens.


bgruening (approved):
Cool!

PR #4053, opened 2026-06-11
Add article on NEMO HPC cluster integration with Pulsar

dSizovs:


mvdbeek:
Really cool that this is already used in production 😍 . Can i ask what version you've deployed ? https://github.com/galaxyproject/pulsar-relay/releases/tag/v0.2.0 contains some (hopefully minor) breaking changes. The easiest way to deploy the upgrade would be to initialize a new redis/valkey backend, which would mean you'd temporarily lose in-flight statuses. If you anticipate a high request volume I can provide a migration script, let me know.

dSizovs:
@mvdbeek Thanks!! 😄 It's 0.1.0 from mvdbeek/pulsar-relay (commit 9fb6c8e), predating the tagged releases. I'd like to upgrade to v0.2.0 now, in the nearest future. Volume is low, so is re-initializing a fresh Valkey backend enough, or do the v0.2.0 breaking changes need the migration script even for a small deployment? Happy to take the script if so. Also, anything besides the Valkey re-init I should watch for in the breaking changes?🤔
Thanks again!

mvdbeek:
> re-initializing a fresh Valkey backend enough

yes, that's enough, the rest should be handled already.

PR #4100, opened 2026-07-01
Updated website text

Sch-Da:
I replaced the main publication with the most recent one and added the ssh subdomain. While doing that, I noticed small inconsistencies, which I fixed.


bgruening (approved):

Sch-Da:
thanks for merging!

PR #4108, opened 2026-07-03
Deprecate toolbox filter documentation

mvdbeek:
## Summary
- marks user-defined toolbox filters as deprecated after the Galaxy 26.1 release series
- clarifies that toolbox filters only hide tools from the toolbox and never enforced hard access restrictions
- points admins to tool panel views for alternate toolbox layouts and TPV/dynamic destinations for hard execution policy

## Testing
- not run; documentation-only change


bgruening:
Seems that we are not deprecating it yet, closing.

PR #4114, opened 2026-07-07
Consistency in naming NFDI in Hub messages

Sch-Da:
As suggested here: https://github.com/usegalaxy-eu/issues/issues/938#issuecomment-4197196396
I removed  NFDI as a funder to only keep the specific NFDI projects that are actual funders from recent blog posts on the hub.
Instead, I added NFDI as a tag if the post did not already have it.
The idea is to differentiate between actual funding (bioimaging, dataplant, etc.) and the general use in an NFDI context, which is now available as a tag now.


bgruening:
Danke

PR #4124, opened 2026-07-09
Upgrade the Astro site to Astro 7

dannon:
Astro 7 is out and dependabot opened #4122 to bump it, but that PR grabs a dozen unrelated majors at once. This does the framework upgrade on its own so it's actually reviewable.

Handles the two v7 breaking changes that matter for us: the new Sätteri markdown pipeline doesn't run remark/rehype plugins, so I opt back into `unified()` (via `@astrojs/markdown-remark`) to keep heading slugs + autolink anchors; and `compressHTML` now defaults to `'jsx'`, which strips inline whitespace, so it's pinned back to `true`. Also bumps tailwind to 4.3 for Vite 8 support and nudges vitest to 4.1. One quirk: Astro 7 backgrounds `astro dev` under AI coding agents, which breaks our dev script and Playwright, so both force `ASTRO_DEV_BACKGROUND=0`.

Left out of scope (separate PRs): citation-js, slugify, lucide, marked, eslint 8->10, linkinator.

Build, unit tests (184), lint/format, and the full Playwright suite (145) all pass.

### DIFF
--- a/astro/package.json
+++ b/astro/package.json
@@ -30,10 +30,11 @@
     "content:lint": "node src/build/normalize-content.mjs --all --check && node src/build/check-dir-names.mjs"
   },
   "dependencies": {
-    "@astrojs/mdx": "^5.0.0",
+    "@astrojs/markdown-remark": "^7.2.1",
+    "@astrojs/mdx": "^7.0.0",
     "@astrojs/rss": "^4.0.17",
     "@astrojs/sitemap": "^3.7.1",
-    "@astrojs/vue": "^6.0.0",
+    "@astrojs/vue": "^7.0.0",
     "@citation-js/core": "^0.6.9",
     "@citation-js/plugin-bibtex": "^0.6.9",
     "@fontsource/atkinson-hyperlegible": "^5.2.8",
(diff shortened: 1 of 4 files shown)


astro/package.json
```diff
@@ -12,7 +12,7 @@
     "generate-redirects": "node src/build/generate-redirects.mjs",
     "generate-404-lookup": "node src/build/generate-404-lookup.mjs",
     "release-guardians:sync-event": "node src/build/release-guardians-sync-event.mjs",
-    "dev": "npm run preprocess && npm run generate-redirects && npm run search-index && npm run generate-404-lookup && (trap 'kill 0' EXIT; node src/build/preprocess.mjs --watch-only & astro dev)",
+    "dev": "npm run preprocess && npm run generate-redirects && npm run search-index && npm run generate-404-lookup && (trap 'kill 0' EXIT; node src/build/preprocess.mjs --watch-only & ASTRO_DEV_BACKGROUND=0 astro dev)",
```
bgruening, line 15: Is that Dev background here needed?

bgruening (approved):
Do we need an issue about the Satteri pipeline?

dannon:
Yep, good call -- opened #4132 to track it. The `unified()` processor is a supported escape hatch for now, but the direction is clearly Sätteri; once we can get heading anchors there (or decide to live without them) we should switch and drop the `@astrojs/markdown-remark` dependency.

astro/package.json
dannon: Yep -- Astro 7 auto-detaches `astro dev` into a background daemon when it detects an AI coding agent. When that happens, the subshell here exits immediately, the `trap 'kill 0'` kills the preprocess watcher, and you're left with an orphaned server. It's a no-op for a human running `npm run dev` in a normal terminal; it only kicks in for agent-driven runs, which is exactly where it breaks. Same reason it's set in the Playwright config -- otherwise Playwright thinks the webServer exited early.

PR #4139, opened 2026-07-20
Upgrade the remaining major dependencies deferred from the Astro 7 PR

dannon:
Follow-up to #4124, taking care of the rest of the majors that dependabot's #4122 tried to land in one shot: linkinator 7, @sindresorhus/slugify 3, marked 18, @citation-js 0.8, lucide 1.x, and eslint 10. One commit per upgrade, each verified independently.

Most were mechanical, verified by diffing generated output before and after: slugify's generated redirects and the citation-js citations pages are byte-identical, and marked 18's only output change is the documented trailing-blank-line trim, which shows up as cosmetic whitespace between tags in three feed files.

Two needed real changes:

- **lucide 1.0** renamed `lucide-vue-next` to `@lucide/vue` (same API, import swap) and dropped brand icons. Content still references `github`/`twitter`/`linkedin` through the mdx `Icon` component, so those three SVGs are vendored from lucide-static 0.575 (ISC) in `brand-icons.ts` with a fallback lookup in `Icon.astro` -- the built site renders identically, verified by comparing unknown-icon spans and brand SVG presence across the full build.
- **eslint 10** drops legacy `.eslintrc` support, so the config moves to `eslint.config.mjs` with the same semantics (same recommended sets, same rule overrides), the plugins move to their current majors (`eslint-plugin-astro` 2, `eslint-plugin-vue` 10), `@typescript-eslint/*` becomes the `typescript-eslint` meta package, and the lint scripts drop `--ext`. The flat config lints exactly the same 260 files the old setup covered. The new recommended rules flagged a handful of real minor issues (dead initializers, required-props-with-defaults) -- fixed in their own commit.

Rounded off with an `npm update` to refresh in-range minors (astro 7.1, lucide 1.25, playwright 1.61, prettier 3.9 -- which reformatted two files, type-level only).

Build, unit tests (191), lint, format check, and the full Playwright suite (145) all pass at the tip.


bgruening (approved):

bgruening:
Thanks

PR #4143, opened 2026-07-20
fix(gcc2026): rename Keynote.jpg to avoid case-only filename collision

itisAliRH:
## Problem

Two **different** images are tracked at paths that differ only in case:

| Path | Size | Referenced by |
|---|---|---|
| `content/images/events/gcc2026/Keynote.jpg` | 56,947 B | `content/news/2026/2026-07-06-gcc2026-recap/index.md` |
| `content/images/events/gcc2026/keynote.jpg` | 102,199 B | `content/events/gcc2026/index.md`, `astro/src/content/events/events--gcc2026.md` |

On case-insensitive filesystems (macOS APFS/HFS+ default, Windows NTFS default) both paths resolve to one file. Consequences for contributors on those platforms:

- Only one of the two images ends up in the working tree; the other is silently absent.
- `git status` reports a permanent, unfixable modification to `keynote.jpg` (its contents are actually `Keynote.jpg`'s blob).
- `git checkout` / branch switches can flip which image is present.

`Keynote.jpg` was added in 5b92312d1 (GCC2026 news post photos) alongside the pre-existing lowercase `keynote.jpg`, so the collision has been present since then.

## Fix

Rename the news-recap image to a distinct, lowercase name and update its single reference:

- `content/images/events/gcc2026/Keynote.jpg` → `content/images/events/gcc2026/keynote-recap.jpg`
- `content/news/2026/2026-07-06-gcc2026-recap/index.md` — `src` updated to match

The event-page image keeps its original `keynote.jpg` path, so neither the event page nor the Astro event content changes.

## Note on the changed URL

`/images/events/gcc2026/Keynote.jpg` will 404 after this merges. It is referenced only from the recap news post, which is updated here, so there is no in-repo breakage. External deep links to that image (if any) would break — flagging in case a redirect is wanted.

## Verification

- Both images now present in the working tree simultaneously.
- `grep -rn "eynote.jpg" content/ astro/` returns only the intended references.
- Clean `git status` on macOS after checkout — no phantom modification.


bgruening:
It's now the second time we see this. Is ther a way we can lint it? Should we enforce all lowercase filenames?

dannon:
Took a crack at linting for this in #4145 -- fails CI on any case-only path collision (this exact `Keynote.jpg`/`keynote.jpg` thing), and blocks newly added asset files with uppercase names. Should catch this kind of thing before it lands.

PR #4145, opened 2026-07-20
Lint for filename case collisions and enforce lowercase on new assets

dannon:
Follow-up to #4143. That was the second time a case-only filename collision (`Keynote.jpg` vs `keynote.jpg`) slipped in, so this adds a lint to catch the class of problem going forward.

`check-filename-case.mjs` joins the `content:lint` pipeline with two checks:

1. **Case-collision (always, repo-wide):** fails if any two tracked paths are equal when lowercased. These collide on case-insensitive filesystems (macOS, Windows) -- only one file lands in the working tree, git reports a permanent phantom modification, and branch switches flip which one is present. The repo is clean today, so this passes.
2. **New-asset lowercase (forward-only):** fails if a PR *adds* an image/media/doc asset with an uppercase basename. It only looks at files added relative to the PR base, so the ~1,600 existing uppercase files are untouched -- no mass rename. The content-lint job checks out at depth 2 and diffs against `HEAD^1` of the PR merge commit (the current base tip), so no full-history fetch is needed.

Includes unit tests and a CONTRIBUTING note. I deliberately didn't try to enforce lowercase on all existing files -- that's a much bigger change and a separate discussion if we want it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+        with:
+          # Full history so the filename-case lint can diff added files
+          # against the PR base ref.
+          fetch-depth: 0
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
@@ -71,6 +75,8 @@ jobs:
           cache-dependency-path: astro/package-lock.json
 
       - name: Lint content
+        env:
+          LINT_BASE_REF: ${{ github.event.pull_request.base.sha }}
         run: cd astro && npm ci && npm run content:lint
 
   unit-tests:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,9 +63,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v7
         with:
-          # Full history so the filename-case lint can diff added files
-          # against the PR base ref.
-          fetch-depth: 0
+          # Depth 2 so the PR merge commit's parents are present: HEAD^1 is
+          # the current base branch tip, which the filename-case lint diffs
+          # added files against.
+          fetch-depth: 2
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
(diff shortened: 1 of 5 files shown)


bgruening (commented):
Thanks

itisAliRH (commented):
Reviewed early since it's a draft; nice idea, and grandfathering old files is the right call. Two spots where the lint passes when it shouldn't:

Renames slip past. `listAddedFiles` uses `git diff --diff-filter=A`, and git detects renames by default, so `git mv keynote.jpg Keynote.jpg` shows up as `R`, not `A`. That's a likely way this bug comes back.

```sh
git(['diff', '--no-renames', '--diff-filter=A', '--name-only', '-z', base])
```

The CLI guard can quietly do nothing.

```javascript
if (import.meta.url === `file://${process.argv[1]}`) {
```

`import.meta.url` is percent-encoded, so if the checkout path has a space or a non-ASCII character, this is false, `main()` never runs, and the script exits 0 with no output. CI goes green having checked nothing.

```javascript
import { pathToFileURL } from 'url';
if (import.meta.url === pathToFileURL(process.argv[1]).href) {
```

Smaller notes:

- `base.sha` is the base tip from the last sync, not the merge-base. An uppercase file merged to main after that point can get blamed on this PR. Rare, but confusing.
- `fetch-depth: 0` pulls every image blob on each lint run. A shallow checkout plus `git fetch --depth=1 origin <base.sha>` gets the same thing cheaper.
- `/[A-Z]/` misses accented capitals like `Ä.jpg`; `/\p{Lu}/u` if you want it tight.

Rest looks good — exit codes, messages, and helper tests all line up with the other lints.

dannon:
Thanks for the thorough review, @itisAliRH -- the rename bypass and the entry-guard bug were both real, nice catches. All five points are addressed in 8fe65bc:

- `listAddedFiles` now diffs with `--no-renames`, so a case-only `git mv` shows up as an add. Verified locally that renaming `call_for_trainers.png` -> `Call_For_Trainers.png` fails the lint now.
- The CLI guard uses `pathToFileURL(process.argv[1]).href`, so it can't silently no-op on paths with spaces.
- Base resolution goes through `merge-base`, and CI now passes `HEAD^1` of the PR merge commit instead of `event.pull_request.base.sha` -- since the checkout is the merge commit, `HEAD^1` is always the *current* base tip, which kills the stale-base problem entirely.
- That also meant `fetch-depth: 0` could drop to `fetch-depth: 2` -- no more full-history fetch on every lint run.
- `/\p{Lu}/u` for the uppercase test, with a test case for `Ä`.

Taking this out of draft -- thanks again!

PR #4152, opened 2026-07-24
Update Earth SIG

Marie59:
Hi all, 

Here an update of the SIG with the latest stuffs done on the earth side. With @annefou we thought it could be nice to merge the climate SIG in the Earth one. So I integrated some classics use case of the climate community in the SIG.
Do you think you could make the climate SIG disappear.

Anne are you okay with the changes ? 

PS: @yvanlebras and @PaulineSGN I noticed the Ecology SIG does not have a page and it only sends back to the subdomain (maybe its on purpose)

### DIFF
--- a/content/community/sig/earth/index.md
+++ b/content/community/sig/earth/index.md
@@ -15,7 +15,7 @@ Whether you are analysing geosphere, biosphere, cryosphere, hydrosphere, and atm
 
 **Tools for earth data analysis are freely available** in the ToolShed which can be installed on any Galaxy server.
 
-If tools are missing or information is not up-to-date in the list, please help us! Contact [Marie](mailto:marie14.josse@gmail.fr) or [Jérôme](mailto:jerome.detoc@ifremer.fr) about it.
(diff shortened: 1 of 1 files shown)


yvanlebras:
THANK YOU Marie! True, we effectively have initiatied an Ecology SIG in the past https://github.com/galaxyproject/galaxy-hub/pull/3398/changes but not sure why it is not linked to the webpage... Maybe I forget something... Thank you, your ping is a good way to go forward, updating it and trying to add the sig page as the link reachable through SIGs pages

PaulineSGN:
> THANK YOU Marie! True, we effectively have initiatied an Ecology SIG in the past https://github.com/galaxyproject/galaxy-hub/pull/3398/changes but not sure why it is not linked to the webpage... Maybe I forget something... Thank you, your ping is a good way to go forward, updating it and trying to add the sig page as the link reachable through SIGs pages

It's a link error on the SIG first page; I will change that.

bgruening (approved):

PR #4161, opened 2026-07-27
Add print stylesheet

### LINKED ISSUE #3544: Add print stylesheet
## Description

Add print-specific styles so articles and documentation print cleanly.

## Tasks

1. Add `@media print` styles to hide navigation, sidebar, footer
2. Ensure content is readable in print (appropriate fonts, sizes)
3. Handle images appropriately (don't break across pages)
4. Add URL display for links when printing
5. Test with browser print preview

## Key considerations

- Hide interactive elements (search, mobile menu toggle)
- Ensure code blocks print with syntax visible
- Handle tables that might be too wide

dannon:
Closes #3544.

Marten's comment on the issue is the real starting point here -- articles don't just print unstyled, they print *wrong*. Repro'd it on the GCC2026 recap page and the print output was dropping content: paragraphs cut off mid-sentence, one heading rendered as nothing but its gold underline, and the "Photo Credit / CC BY-SA" block was missing from the PDF text layer entirely.

Three things were causing it:

- The nested flex shells in `BaseLayout` (`.site-shell` > `.site-main` > `.site-content`) fragment badly across printed pages in Chrome.
- `overflow: hidden` on the outer wrapper clips everything past page one. `overflow-x: hidden` on `html`/`body` does the same thing indirectly, since a non-visible value on one axis computes the other to `auto` and makes the root a scroll container.
- The mobile header is `position: fixed`, and Chrome repaints fixed elements on *every* printed page, over whatever is underneath. Print renders at ~816px, below the `lg:` breakpoint, so it's the mobile layout that prints -- which is why nobody noticed the header was involved.

`print.css` unwinds all three back to plain block flow and hides the nav, sidebar, footer, TOC and other interactive bits.

There's a second class of bug worth calling out, because it only shows up with **Background graphics off** -- which is the default in Chrome's print dialog, and is how Marten's screenshot was taken. The page header, table header cells and code blocks are all light-text-on-dark-background, so with backgrounds suppressed they render white-on-white and disappear. Article titles were literally invisible. Those now use borders and dark text rather than depending on a background, so they read the same either way.

Also in here, from the issue's task list: link targets print after the link text (relative ones expanded to absolute), images and tables get `break-inside: avoid`, long code lines wrap instead of scrolling off the page, and `thead` repeats when a table spans pages. URLs are deliberately suppressed inside table cells -- the release notes are full of link-dense tables and a URL in every cell squeezed the columns until dates were wrapping as "Octob er 3rd".

## On the diff

The changes to existing files are only class-name hooks (`site-shell`, `site-mobile-header`, `site-footer`, `site-sidebar`, etc.) plus one `@import`. Everything else is new. All rules live inside `@media print`, so screen rendering is untouched.

## Testing

`tests/print.spec.ts` asserts computed styles under print media emulation rather than comparing pixels, so it shouldn't churn as content changes. 13 tests; 9 of them fail with the stylesheet reverted, so they're actually holding the fix down.

Verified end-to-end against the production build (not just dev): the recap page goes 16 pages -> 11, and the content that was being dropped is back. Existing suite is green (191 unit tests, lint, prettier).

Worth a look at the rendered before/after if you're reviewing -- the difference on the `#training` section Marten linked is pretty stark.

### DIFF
--- a/astro/src/components/layout/ContentWithToc.astro
+++ b/astro/src/components/layout/ContentWithToc.astro
@@ -40,7 +40,7 @@ const showToc = autotoc && filteredHeadings.length >= 2;
 }
 
 {/* Content with optional TOC sidebar */}
-<div class:list={['px-6 sm:px-8 pt-6', showToc && 'lg:flex lg:gap-8']}>
+<div class:list={['content-with-toc px-6 sm:px-8 pt-6', showToc && 'lg:flex lg:gap-8']}>
   {/* Main Content */}
   <div class:list={['prose prose-lg prose-galaxy max-w-none', showToc && 'lg:flex-1 lg:min-w-0']}>
     <slot />
--- a/astro/src/components/layout/Sidebar.astro
+++ b/astro/src/components/layout/Sidebar.astro
@@ -14,7 +14,9 @@ interface Props {
 const { navbar, initialSubsite } = Astro.props as Props;
 ---
 
-<aside class="hidden lg:flex flex-col w-64 bg-galaxy-dark text-white fixed left-0 top-0 h-screen overflow-y-auto z-40">
+<aside
+  class="site-sidebar hidden lg:flex flex-col w-64 bg-galaxy-dark text-white fixed left-0 top-0 h-screen overflow-y-auto z-40"
+>
   <!-- Logo Section -->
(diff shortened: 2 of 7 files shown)


martenson:
thanks @dannon

PR #4168, opened 2026-07-30
The EU Galaxy project and the Freiburg Galaxy team's website are now served fully from the Hub

bgruening:
Announcement post, but also advertisement for other Galaxy servers to join us.

### DIFF
--- a/content/news/2026/2026-07-30-eu-website-on-galaxy-hub/index.md
+++ b/content/news/2026/2026-07-30-eu-website-on-galaxy-hub/index.md
@@ -1,0 +1,66 @@
+---
+title: "The European Galaxy Project Website Is Now Fully Served by the Galaxy Hub"
+date: "2026-07-30"
+tease: "After ten years on a Jekyll-on-GitHub setup, galaxyproject.eu is now completely powered by the Galaxy Hub — a fast, metadata-driven home that any Galaxy community can reuse."
+subsites: [global, eu, freiburg]
+main_subsite: eu
+tags: [eu, freiburg, community, infrastructure]
+contributions:
+  authorship:
+    - bgruening
+  funding:
+    - eu
+    - deNBI
+---
+
+# The European Galaxy Project Website Is Now Fully Served by the Galaxy Hub
+
+The European Galaxy project has run its team website at [galaxyproject.eu](https://galaxyproject.eu/) for a decade. It started life ten years ago as a Jekyll-powered site hosted on GitHub — a pragmatic choice at the time, and one that served the community well for years. A few years back we began migrating parts of it (posts and news) over to the Galaxy Hub, but a handful of features, most notably the [galaxyproject.eu/people](https://galaxyproject.eu/people) page, the [galaxyproject.eu/citations](https://galaxyproject.eu/citations) page and several other pages for Galaxy subdomains, had stubbornly remained on the old infrastructure.
+
+**As of today, the last bits have moved over.** The European Galaxy project website is now completely served by the [Galaxy Hub](https://galaxyproject.org).
+
+This is more than a hosting change. With the new Hub we have a powerful, fast, metadata-driven website that can act as the shared home for many Galaxy servers, and the European community is the first to make full use of it.
+In this post we want to show off a few of the features that let you present your own Galaxy community, share it with the rest of the ecosystem, and participate in the global view of who builds and uses Galaxy.
+
+## One Hub, many homes: subsites for every Galaxy server/community
+
+Every Galaxy server can have its own space inside the Hub. The European Galaxy server (hosted in Freiburg) maintains [galaxyproject.org/freiburg](https://galaxyproject.org/freiburg) and [galaxyproject.org/eu](https://galaxyproject.org/eu), which is also reachable via [galaxyproject.eu](https://galaxyproject.eu/). As the European Galaxy community we look after several national Galaxy servers, and each of them can have its own subsite in the same way.
+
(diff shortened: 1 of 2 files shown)


B0r1sD:
This is a cool change that definitely makes lots of Galaxy meta topics more interconnected, as well as make the center panels of the different instances more streamlined! I would like to share some points that made us refrain in implementing it currently at usegalaxy.be, as requested:
- It's nice one is able to scope the hub page to a certain region, so it shows which people etc. are attached to your region. But if you click a certain 'wrong' subpage - one that shifts your regional view to 'Global', you are just browsing the general hub site inside the EU instance which feels off. I feel like it should be limited to that page only, no browsing around possible for the user.
- For this installed tools view, the sidebars felt a bit too nested: 
  afbeelding
  this in combination with the behavior described above, the request is therefor if it would be possible to make that sidebar inaccessible? Or perhaps adding a bare webpage for tools (and potentially other links) could solve this?
 - News and Events do open a new tab instead of showing the (bare) webpage inside the instance. Is this a deliberate choice or are the bare webpages for these just more complete then https://galaxyproject.org/eu/events/ and https://galaxyproject.org/eu/news/?
 
 I might add one or more additional points after input from my colleagues. Thanks a lot for the initial example!

bgruening:
>     * It's nice one is able to scope the hub page to a certain region, so it shows which people etc. are attached to your region. But if you click a certain 'wrong' subpage - one that shifts your regional view to 'Global', you are just browsing the general hub site inside the EU instance which feels off. I feel like it should be limited to that page only, no browsing around possible for the user.

If I understand your problem correct this is a bug and you should fill an issue. The sidebare is configureable per subpage, so this is up to Belgium to decide where certains links point to.
 
>     * For this installed tools view, the sidebars felt a bit too nested:
>       afbeelding
>       this in combination with the behavior described above, the request is therefor if it would be possible to make that sidebar inaccessible? Or perhaps adding a bare webpage for tools (and potentially other links) could solve this?

Uha, how do you get there?
Two solutions, either we make sure all links are opening in a new window, or as you pointed out the tool page can be as well a /bare/ side.

>     * News and Events do open a new tab instead of showing the (bare) webpage inside the instance. Is this a deliberate choice or are the bare webpages for these just more complete then https://galaxyproject.org/eu/events/ and https://galaxyproject.org/eu/news/?

Thats a deliberate choice, its an iframe and we want to inform users, get some teasers out, but if you want to browse the hub, use a different tab.

bedroesb:
Hey Bjorn! Since I was involved in the design of our current landing page, I got dragged into looking into this new solution. I got a couple of questions :)
- If I understood correctly the custom landing page of eu is hosted through: https://github.com/galaxyproject/galaxy-hub/blob/main/content/bare/eu/usegalaxy/main/index.md?plain=1. It is an interesting to me that Iframes are used for components that live in the same repo: https://github.com/galaxyproject/galaxy-hub/blob/main/astro/src/pages/bare/eu/latest/news.astro Could this not have been an import? The current implementation causes a 3th iframe nested layer in the middle panel
- I like the possibility to still have a custom landing page like EU has and potentially linking to the fully rendered list of things in a new tab, like: https://galaxyproject.org/eu/news/, but as Boris mentions in the comment above, it is not clear enough when people get switched from (in this case) EU specific content, to Global content. A good first step I think would be to move the instance selector more visible on top, so it is clear that the listed sidebar items are linked to that selected instance. Could this be an option? (I felt I had to end with a question in a questions list :D ) 
- Why does https://galaxyproject.org/bare/eu/news/ exist when https://galaxyproject.org/eu/news/ exists?

bgruening:
> Hey Bjorn! Since I was involved in the design of our current landing page, I got dragged into looking into this new solution. I got a couple of questions :)
> 
>     * If I understood correctly the custom landing page of eu is hosted through: https://github.com/galaxyproject/galaxy-hub/blob/main/content/bare/eu/usegalaxy/main/index.md?plain=1. It is an interesting to me that Iframes are used for components that live in the same repo: https://github.com/galaxyproject/galaxy-hub/blob/main/astro/src/pages/bare/eu/latest/news.astro Could this not have been an import? The current implementation causes a 3th iframe nested layer in the middle panel

At the moment It's for demonstration. There is no technical necessity here. Its answer your question 3 below more or less ;)

It could be a simple import `<LatestNews subsite="eu" />` or something similar. What I would need to ask @itisAliRH is if this is then importing more CSS and other artifacts that we want.

Those pages are meant to be embedded in other websites. For example into usegalaxy.eu ... so everyone that decides to use those pages are free to add whatever they want, iframes or not. We just want to make it easy to compose relevant content for people. One example are the subsites/sub-pages like earth-science.usegalaxy.eu or ssh.usegalaxy.eu they compose their own bare-pages to have them embedded later.

>     * I like the possibility to still have a custom landing page like EU has and potentially linking to the fully rendered list of things in a new tab, like: https://galaxyproject.org/eu/news/, but as Boris mentions in the comment above, it is not clear enough when people get switched from (in this case) EU specific content, to Global content. A good first step I think would be to move the instance selector more visible on top, so it is clear that the listed sidebar items are linked to that selected instance. Could this be an option? (I felt I had to end with a question in a questions list :D )

Everything is an option :)
I don't really understand how @B0r1sD got to this screenshot, or what he clicked, but I assume there is somewhere still a link that is not opening in a new window. The sidebar should never have been shown inside the usegalaxy.eu website. The bare-pages do exactly this, they do not render anything specific to the Hub.

It should look like this:

grafik


>     * Why does https://galaxyproject.org/bare/eu/news/ exist when https://galaxyproject.org/eu/news/ exists?

One is a `bare` side, no decorations, no sidebar, no Hub. You can take it and embed it into any website you want. E.g. on the ELIXIR website of the Galaxy community. Or on usegalaxy.eu we have a webhook that displays all events / news etc in 1 out of 10 tool submissions.

The other one is the normal Hub website view, with decorations, Hub style, meant to be consumed from the Hub.

bgruening:
Please lets keep the discussion going, or create an issue (or multiple one), we can also jump into a meeting. I just need this news post now to inform the other sub-sites etc ...

B0r1sD:
> Please lets keep the discussion going, or create an issue (or multiple one), we can also jump into a meeting. I just need this news post now to inform the other sub-sites etc ...

No worries, I'll open an issue for this screenshot and another related bug I saw and how you can recreate it!

PR #4178, opened 2026-08-06
add a draft of user facing UDT docs

scottcain:
While this all *feels* right, please be aware that I used claude to write it; I'd really appreciate a dev or two taking a look at this to verify it.


content/tools/user-defined-tools/index.md
```diff
@@ -0,0 +1,86 @@
+---
+title: User-Defined Tools
+tease: "Write and run your own simple tools from the Galaxy interface — no server access, no admin installation, no Tool Shed."
+---
+
+*Beta feature, available starting with Galaxy 25.0*
```
natefoo: Not sure in what release they matured to the point where we're reasonable comfortable with them but I probably wouldn't recommend running them on 25.0

content/tools/user-defined-tools/index.md
```diff
@@ -0,0 +1,86 @@
...
+
+User-Defined Tools (UDTs) let regular Galaxy users write and run their own simple tools directly from the Galaxy interface — no server access, no admin installation, and no waiting for a tool to be added to a Tool Shed. If you've ever needed a quick script to concatenate files, reformat a table, or glue together a couple of steps in a workflow, and didn't want to file a request with your Galaxy admin, UDTs are for you.
```
natefoo: Concatenate files is maybe not the best example here since we have a tool that does that so you should never need a UDT just to do that. ;)

content/tools/user-defined-tools/index.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Sharing tools
+
+UDTs are private to the user who created them by default. If you embed a UDT in a workflow and share that workflow, anyone who imports the workflow automatically gets a copy of the tool in their own account — they don't need to have created it themselves or have tool-creation permission.
+
+An administrator can also export a UDT to disk and install it like a regular tool, making it available instance-wide.
+
+## Security considerations
+
+UDTs carry similar risks to [Interactive Tools](https://training.galaxyproject.org/training-material/topics/admin/tutorials/interactive-tools/tutorial.html) — both let users run their own code inside containers on your infrastructure. If you're an admin enabling this feature, we strongly recommend reading the [guidance on securing Interactive Tools](https://training.galaxyproject.org/training-material/topics/admin/tutorials/interactive-tools/tutorial.html#securing-interactive-tools) and configuring your job runner to isolate mounts and restrict network access for these jobs, as shown in this [example Pulsar configuration](https://github.com/galaxyproject/galaxy/blob/dev/test/integration/embedded_pulsar_job_conf.yml).
```
natefoo: Might also point out that for isolation purposes you can map UDTs directly in TPV because they are automatically tagged with the `accept` tag `tool_type_user_defined` and all destinations are automatically tagged with a `reject` tag of the same name, so you have to explicitly tag UDT destinations with `accept` or `require` tag `tool_type_user_defined`. We strongly recommend running UDTs via Pulsar on compute that is logically isolated from the rest of Galaxy and its data.

natefoo (commented):
This is relatively accurate, see comments inline. The other other thing I'd maybe like to see is some reassurance that the existing Galaxy goals of reproducibility and collaboration remain important to us and that people should continue to use "real" tools as much as possible so that workflows and so forth can continue to be shared and published.

content/tools/user-defined-tools/index.md
scottcain: I'll be honest--even though claude added the TPV verbiage, I have no idea what any of it means :-/

natefoo: This was bad advice on my part, documenting it in a news release is not the right place. Just have it link to where it's documented in TPV: https://total-perspective-vortex.readthedocs.io/en/latest/topics/advanced_topics.html#auto-injected-tool-type-tags

scottcain: @natefoo is this sufficient (at the end of the previous paragraph, and remove the newly added paragraph)?
```
For more information on the security concerns and ways to address them, see 
the [Total Perspective Vantage (TPV) documentation](https://total-perspective-vortex.readthedocs.io/en/latest/topics/advanced_topics.html#auto-injected-tool-type-tags)
```

natefoo: So that should be Vortex not Vantage, but I forgot about this until now so it's a bit too late :sweat_smile:

PR #4186, opened 2026-08-10
fix(eu): stop bare pages nesting the Hub and Galaxy inside the embedding iframe

### LINKED ISSUE #4183: Hub integration into the Galaxy landing page results in a (broken) sidebar replication
Xref where this has been discussed in more detail: https://github.com/galaxyproject/galaxy-hub/pull/4168

**Bug**

See screenshot for double activity bar, that doesn't respond when clicking a tool. 
Image

**How to replicate**

1. browse to https://usegalaxy.eu/
2. Click 'Browse installed tools'

Image

4. Click any tool (where the presence of the sidebar of the hub website should not be present imo)

Image

5. Results in nested, broken Galaxy EU instance where you can not click a tool (see first picture of this issue)

**Expected behavior**
The link to the tool should open in the actual Galaxy instance.

itisAliRH:
Fixes #4183, closes https://github.com/usegalaxy-eu/issues/issues/1027.

Bare pages are embedded by the Galaxy servers they describe (usegalaxy.eu loads /bare/eu/usegalaxy/main/ in its welcome frame). Links in that content had no target, so:

- "Browse installed tools" loaded the full Hub, sidebar included, inside the usegalaxy.eu iframe;
- Clicking a tool there loaded a whole Galaxy inside that same iframe — double activity bar, tool panel unresponsive.

Changes:
- BareArticleLayout now defaults every link that leaves the page to `target="_blank"` (in-page anchors and author-set targets untouched), so no Hub page can ever render inside an embedding instance;
- tool links are rendered with `target="_top"` so a tool opens in the Galaxy instance itself — unframed (galaxyproject.org/eu/tools/) that behaves exactly as before;
- The two UseGalaxy.eu links in the EU lead section are pinned to `_top`.

Verified with a cross-origin iframe host against a local build: the embed's iframe URL is unchanged after clicking "Browse installed tools" (opens a new tab), and clicking a tool navigates the top window to `https://usegalaxy.eu/root?tool_id=<tool_id>` with the iframe gone.


bgruening (approved):

astro/src/utils/bare-links.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Link target defaults for bare pages.
+ *
+ * Bare pages are embedded as iframes by the Galaxy servers they describe:
+ * usegalaxy.eu loads /bare/eu/usegalaxy/main/ inside its welcome frame. A link
+ * without a target navigates that iframe, so the Hub — sidebar, navbar and all —
+ * ends up rendered inside the server, and a link back to the server renders a
+ * Galaxy inside Galaxy. Neither is ever wanted from an embed, so every link that
```
B0r1sD, line 8: Thanks @itisAliRH! Now every link opens the hub in a new tab which mitigates the issue, and I noticed this comment section does hint towards the possibility of providing specific targets for certain links in case the embedded hub is still the desired behavior. I'll make sure to test this whenever we start integrating the hub at Galaxy BE as this is likely how we would implement it in most cases.

PR #4219, opened 2026-08-24
Add Tools page for US subsite as well using a weekly workflow

ahmedhamidawan:
Uses the exact same approach and command as `/eu/tools/` to add a new `/us/tools/` page as well.

Fixes the tools component of https://github.com/galaxyproject/galaxy-hub/issues/4216

CC: @mschatz

### DIFF
--- a/.github/workflows/update-tools.yml
+++ b/.github/workflows/update-tools.yml
@@ -1,4 +1,4 @@
-name: Update EU Tools Report
+name: Update Tools Report
 on:
   workflow_dispatch:
   schedule:
@@ -12,6 +12,19 @@ jobs:
   update-tools:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'galaxyproject'
+    strategy:
+      matrix:
+        include:
+          - id: eu
+            label: EU
+            server: https://usegalaxy.eu
+            name: 'European Galaxy'
+            output: content/eu/tools/index.md
+          - id: us
+            label: US
+            server: https://usegalaxy.org
+            name: 'US Galaxy'
+            output: content/us/tools/index.md
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -32,14 +45,14 @@ jobs:
         run: pip install requests
 
       - name: Run update script
-        run: python scripts/update-tools.py --server https://usegalaxy.eu --name "European Galaxy" --output content/eu/tools/index.md
+        run: python scripts/update-tools.py --server ${{ matrix.server }} --name "${{ matrix.name }}" --output ${{ matrix.output }}
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v8
         with:
           token: ${{ steps.app-token.outputs.token }}
-          commit-message: 'chore(eu): update tools report'
-          title: 'Update EU Tools Report'
-          body: 'Auto-generated PR with the updated inventory of tools available at [Galaxy EU](https://usegalaxy.eu).'
-          branch: update-eu-tools-report
+          commit-message: 'chore(${{ matrix.id }}): update tools report'
+          title: 'Update ${{ matrix.label }} Tools Report'
+          body: 'Auto-generated PR with the updated inventory of tools available at [Galaxy ${{ matrix.label }}](${{ matrix.server }}).'
+          branch: update-${{ matrix.id }}-tools-report
           delete-branch: true
--- a/.github/workflows/update-tools.yml
+++ b/.github/workflows/update-tools.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'galaxyproject'
     strategy:
+      fail-fast: false
       matrix:
         include:
           - id: eu
@@ -23,7 +24,7 @@ jobs:
           - id: us
             label: US
             server: https://usegalaxy.org
-            name: 'US Galaxy'
+            name: 'Galaxy US'
             output: content/us/tools/index.md
     steps:
       - name: Checkout
@@ -45,7 +46,7 @@ jobs:
         run: pip install requests
 
       - name: Run update script
-        run: python scripts/update-tools.py --server ${{ matrix.server }} --name "${{ matrix.name }}" --output ${{ matrix.output }}
+        run: python scripts/update-tools.py --server "${{ matrix.server }}" --name "${{ matrix.name }}" --output "${{ matrix.output }}"
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v8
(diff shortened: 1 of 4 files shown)


ahmedhamidawan:
@itisAliRH I hope the workflow I've added makes sense. I tried to replicate what you did in https://github.com/galaxyproject/galaxy-hub/pull/3932. 

The page is generated correctly as expected and tested locally.

.github/workflows/update-tools.yml
bgruening, line 1: Should we put those into one CI run / file? We probably don't X CI jobs for X servers?

bgruening (commented):
Cool, thanks @ahmedhamidawan!

.github/workflows/update-tools.yml
ahmedhamidawan: That makes complete sense, maybe I can extend the same workflow...

ahmedhamidawan: Done! I hope the new workflow looks good!

 @itisAliRH

.github/workflows/update-tools.yml
```diff
@@ -12,6 +12,19 @@ jobs:
...
+          - id: eu
+            label: EU
+            server: https://usegalaxy.eu
+            name: 'European Galaxy'
+            output: content/eu/tools/index.md
+          - id: us
+            label: US
+            server: https://usegalaxy.org
+            name: 'US Galaxy'
+            output: content/us/tools/index.md
```
itisAliRH, line 28: `content/us/tools/` does not exist in the repo, and `scripts/update-tools.py` writes the report with a plain `open(args.output, "w")` without creating the parent directory; so this matrix entry will fail with `FileNotFoundError` on its very first run.

The EU page only works because `content/eu/tools/index.md` was committed manually first.

Either commit a stub content/us/tools/index.md in this PR, or make the script create the directory:

```
import os
os.makedirs(os.path.dirname(args.output) or ".", exist_ok=True)
```

.github/workflows/update-tools.yml
```diff
@@ -12,6 +12,19 @@ jobs:
   update-tools:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'galaxyproject'
+    strategy:
```
itisAliRH, line 15: Please add `fail-fast: false`. With the matrix default `(fail-fast: true)`, a failure in either job cancels the other job, so a US-side breakage would silently take down the weekly EU tools PR that works today.

```suggestion
    strategy:
      fail-fast: false
```

.github/workflows/update-tools.yml
```diff
@@ -12,6 +12,19 @@ jobs:
...
+        include:
+          - id: eu
+            label: EU
+            server: https://usegalaxy.eu
+            name: 'European Galaxy'
+            output: content/eu/tools/index.md
+          - id: us
+            label: US
+            server: https://usegalaxy.org
+            name: 'US Galaxy'
```
itisAliRH: `US Galaxy` renders the page title as `US Galaxy Tools`. The hub calls this subsite `Galaxy US` (`content/us/main.md`) and the instance "usegalaxy.org (Main)" (`content/use/usegalaxy-org/index.md`). `Galaxy US` would read more consistently.


```suggestion
            name: 'Galaxy US'
```

.github/workflows/update-tools.yml
```diff
@@ -32,14 +45,14 @@ jobs:
         run: pip install requests
 
       - name: Run update script
-        run: python scripts/update-tools.py --server https://usegalaxy.eu --name "European Galaxy" --output content/eu/tools/index.md
+        run: python scripts/update-tools.py --server ${{ matrix.server }} --name "${{ matrix.name }}" --output ${{ matrix.output }}
```
itisAliRH: `--server` and `--output` are unquoted while `--name` is quoted. The current values are safe, but quoting all three keeps the line robust if a future matrix entry ever contains a space:

```suggestion
        run: python scripts/update-tools.py --server "${{ matrix.server }}" --name "${{ matrix.name }}" --output "${{ matrix.output }}"
```

itisAliRH (changes requested):
Thanks @ahmedhamidawan!

Reviewed and found two blocking items inline (missing `content/us/tools/` directory and `fail-fast`), plus two nits.


One item I couldn't comment on inline, since the file isn't in this diff: [content/us/navbar.yml](https://github.com/galaxyproject/galaxy-hub/blob/main/content/us/navbar.yml) has no Tools entry, while [content/eu/navbar.yml](https://github.com/galaxyproject/galaxy-hub/blob/main/content/eu/navbar.yml) does, so the new page would build but wouldn't be linked from the US subsite. The US navbar uses `relativeTo: subsite`, so it would be:

    - label: Tools
      relativeTo: subsite
      target: tools/

itisAliRH:
Also minor: the comment at the top of [.github/workflows/release-guardians-refresh.yml](https://github.com/galaxyproject/galaxy-hub/blob/main/.github/workflows/release-guardians-refresh.yml) still refers to update-eu-tools after this rename.

ahmedhamidawan:
@itisAliRH Great catches! Sorry, I had several of these things already but I made the dumb mistake when I dropped that first commit by rebasing 🤦‍♂️ .

I'll incorporate all of these fixes

.github/workflows/update-tools.yml
ahmedhamidawan: I did initially (before I rebased and removed that first commit 🤦 ) manually add an `index.md` at the new location, doing that again.

ahmedhamidawan:
@itisAliRH , should be good now hopefully!

itisAliRH (approved):
Thanks, @ahmedhamidawan! No worries about the rebase, that happens to all of us! 🙂

Just two small leftovers:

1. The committed stub was generated before the name change, so it still carries the old value. The next scheduled run will rewrite it to 'Galaxy US Tools', but until then, the live page title won't match what the workflow produces. 

2. Still open from my earlier comment: the header comment in .github/workflows/release-guardians-refresh.yml references update-eu-tools, which no longer exists after this rename.

Both are trivial!🚀

content/us/tools/index.md
```diff
@@ -0,0 +1,2950 @@
+---
+# THIS FILE IS AUTO-GENERATED. DO NOT EDIT MANUALLY.
+# To update, run: python3 scripts/update-tools.py --server https://usegalaxy.org --name "US Galaxy" --output content/us/tools/index.md
+title: US Galaxy Tools
```
itisAliRH:
```suggestion
# To update, run: python3 scripts/update-tools.py --server https://usegalaxy.org --name "Galaxy US" --output content/us/tools/index.md
title: Galaxy US Tools
```

ahmedhamidawan:
@itisAliRH oof! Sorry 😬 .... Should be good now, I pray 🙏

PR #4223, opened 2026-08-26
Add getting-started page for agentic Galaxy development

nekrut:
Adds `/tools/agentic-development/` — a short getting-started guide for driving Galaxy from an AI agent (Orbit, Claude Code, Codex CLI, Gemini CLI) through the Galaxy MCP server, prototyping analysis steps as user-defined tools, and maturing that work toward the IUC and IWC.

Opening as a **draft** — the underlying components are alpha/beta and moving quickly, and I would like eyes on the accuracy before this goes live.

## What it covers

Eight sections, four figures, aimed at a developer who has never used Galaxy and wants to be productive in a day: the two-pane working model, what you need, agent-driven setup, the prototyping loop, a reviewer's checklist, where the work goes next, a dated status box, and further reading.

It complements rather than duplicates [`/tools/orbit/`](https://galaxyproject.org/tools/orbit/) and [`/tools/user-defined-tools/`](https://galaxyproject.org/tools/user-defined-tools/), linking to both for detail.

## Setup is written as prompts, not config files

The agent has a shell, so it registers its own MCP server and clones the skills repo. One prompt replaces four per-harness JSON/TOML configurations. Readers are told to `export GALAXY_API_KEY` and have the agent reference the *variable* — pasting a key into the conversation puts it in the transcript and sends it to the model provider.

## Instructions were executed before writing

Every command was run. Corrections that came out of that, and which reviewers may want to check independently:

- **`GALAXY_MCP_EXCLUDE_TAGS=write` is silently ignored by the Node server.** On the Python server (`uvx galaxy-mcp`) it correctly takes the catalog from 45 tools to 30. On `npx @galaxyproject/galaxy-mcp` the count is 37 before and after, still exposing `run_tool`, `create_user_tool`, `delete_user_tool`, `invoke_workflow` and `upload_file`. The page recommends the Python server for read-only mode and says why. This may be worth an issue against `galaxy-mcp`.
- **`connect` is itself tagged `write`**, so read-only mode removes it — while the unauthenticated error still says "run connect()" and never mentions the env vars. The page pre-empts this.
- **`search_iwc_workflows` needs no Galaxy credentials**, which makes it a usable smoke test before an API key is working.
- **The `udt-authoring` skill ships `scripts/validate.py`**, which runs Galaxy's real `UserToolSource` schema and linter offline with no server, catching the most common agent mistake (an output declaring neither `from_work_dir` nor `discover_datasets`).
- **The usegalaxy.org menu entry is "Manage Galaxy API Key"**; the page it opens is titled "Manage API Key". The GTN FAQ says "Manage API key".

## Points a reviewer may want to weigh in on

- **User-defined tool availability** is stated as requiring *both* `enable_beta_tool_formats` *and* a per-account `user_tool_execute` role. The flag is set on usegalaxy.org and usegalaxy.eu, but role grants live in the database and are not publicly verifiable, so the page teaches readers to check rather than asserting availability. Someone with operational knowledge should confirm this is the right framing.
- **The Claude Code, Codex and Gemini configurations are derived** from each client's own documentation — `galaxy-mcp` publishes a Claude Desktop config only. Claude Code and Codex were verified by execution; **Gemini CLI was not installed on the test machine and is marked untested**.
- **No UDT-to-XML conversion procedure is given**, deliberately. Planemo has no UDT support and the graduated promotion path presented at GCC2026 is a direction of travel rather than shipped functionality, so the page names IUC and IWC as destinations and notes an agent can do the rewrite, without inventing an unofficial recipe.
- **Galaxy 26.1's in-tree MCP server** is mentioned as forward-looking only. `/api/mcp` returned 404 on usegalaxy.org, .eu and .org.au when this was written.

## Status box

The page carries a dated status callout, so it is clear what was true on 2026-08-26 and what will need revision as `galaxy-mcp` (alpha), Orbit (beta), user-defined tools (beta) and the in-tree MCP server (experimental) move.

## Intended destination

This material is aimed at the GTN eventually; it lives on the hub for now.

https://claude.ai/code/session_018C8SdBrwFZddd8n6ypzGVG


nekrut:
It needs to be carefully reviewed by human before flipped from draft.

PR #4234, opened 2026-09-02
Rename key 'kysrpex' to 'domgz' in people.yaml

Sch-Da:
@domgz, I just noticed that your old GitHub handle was still on the website, and I adjusted it so your avatar will also show there. Please double-check :)


domgz (approved):
❤️

PR #4283, opened 2026-09-17
Add Canada to Community dropdown

ccoulombe:
Added Canada dropdown menu with external redirect


ccoulombe:
closes https://github.com/usegalaxy-ca/usegalaxy.ca/issues/277

jdavcs:
Happy to merge this. 

However, as a consideration, don't we want these to be local pages? The idea, as far as I remember, was an integrated approach, so that news and events and other structured content can be pulled (eventually) from all community pages and displayed centrally. This, of course, breaks down with a redirect. Furthermore, I think a unified look for all community pages underscores that all these instances are part of the same close-knit community, or family. That said, @ccoulombe brought up a legitimate concern: if a community page is local to the hub, editing it has to go through a PR, which introduces a delay, which is less than convenient for such content. Merging this now, but leaving this up for discussion. @bgruening

bgruening:
The Hub is treated as a wiki page, you don't need to wait for a review, especially not if you edit your own local pages.

I would argue the build should still pass :) But that is also true for other pages, I assume :)

bgruening:
It would be great if usegalaxy.ca would be a full subsite of the Hub, so we could work together on news/events, share them, and maintain one central Community Hub.

ccoulombe:
I cannot edit the hub as I do not have write permissions, so this is a blocker for frequent edits on a subsite. (I am not necessarily asking write permissions here, just pointing it out.)

That said, a subsite with a front matter for mainly news and events, which we do not have on our actual site (only docs and policies) would be a one-off I think.

bgruening:
> After a PR or two, maintainers will add you to the Editors group so you can commit directly.

From https://github.com/galaxyproject/galaxy-hub/blob/main/CONTRIBUTING.md :)

So everyone is welcome to contribute to the community hub and get commit access.

ccoulombe:
Ah! This is the second PR :)

PR #4285, opened 2026-09-17
Add "Galaxy for AI Coding Agents" page (/tools/ai-agents/)

nekrut:
Adds a single page, `/tools/ai-agents/`, for using Galaxy from AI coding agents, backed by the new [galaxyproject/agentic-plugins](https://github.com/galaxyproject/agentic-plugins) repository (galaxy-mcp server config + mirrored galaxy-skills and Foundry skills, packaged for each harness).

## Landing page: `/tools/ai-agents/`

Styled after [galaxyproject.github.io/loom](https://galaxyproject.github.io/loom/): a full-bleed dark hero and a **live animated replay** of installing the plugins into each harness and then talking to Galaxy through it.

- `AgentShells` MDX component (`astro/src/components/mdx/AgentShells.astro`, runtime in `astro/src/components/agent-shells/`): hero with grid/filament SVG and grain, harness chips, a tab strip and one window whose chrome switches per harness (terminal, or a light chat window for Claude Desktop). Scenes type the real install commands, show the URL/key prompt, verify with `/mcp`, then run a short conversation with MCP tool cards (`get_user`, `create_history`, `run_tool`, …). Scenes auto-advance and loop.
- Controls: pause, replay, copy install commands, "Set up <harness>". Autoplay starts when scrolled into view, pauses when hidden or off screen, and honours `prefers-reduced-motion` by rendering the final state statically. Tabs are keyboard-navigable (arrow keys, Home/End).
- Scene data lives in `agent-shells/scenes.ts` and mirrors the guide pages; tool names were checked against the galaxy-mcp server. Cursor's scene is written but held back from the page for now (`heldBack` export).
- Below the hero: three pillars (Galaxy connection, curated skills, Workflow Foundry), a "Pick your agent" card grid, then the prose (before you start, example prompts, get help).

Layout plumbing, opt-in per page:

- `skip_title_render: true` (already in the schema) is now honoured: the layout PageHeader is skipped so the hero supplies the only h1.
- new `full_bleed: true`: BaseLayout drops the outer content padding and the article loses its max width, so the page runs edge to edge. Documented in `astro/README.md`.
- `CodeCopy` (`astro/src/components/content/CodeCopy.astro`, included by ArticleLayout): every fenced code block in article prose gets a **Copy** button (clipboard API with fallback, "Copied" state, hidden in print). Site-wide, so the install blocks on the guides and every other article page get it.

## Review pass

A 191-agent multi-lens review (runtime, rendering safety, a11y, CSS, content accuracy, build integration, tests, perf, plus a completeness critic) confirmed 61 findings; all are fixed here except a CI drift check for the exporter (tracked upstream). Highlights:

- **Guides lost their intros**: `autotoc: true` made the hub's TOC injector delete everything before the first `##` on every generated guide (the API-key page no longer said how to get a key). Fixed in the exporter (`autotoc: false`), pages regenerated.
- **Generated pages now carry `generated_from`**, and "Edit on GitHub" becomes "Edit the source on GitHub" pointing at `agentic-plugins/docs/`, so the next export cannot silently revert hub edits.
- **A11y**: live summary inside the tabpanel, focusable transcript region, `aria-labelledby` follows the active tab, no `aria-pressed` on the play/pause control, pillars section has an h2, auto-advance waits while keyboard focus is in the tab strip, desktop-window text meets 4.5:1.
- **Runtime**: switching tabs while paused stays paused (shows that harness's final state), autostart honours a prior pause, spinner frames come from the pause-aware loop (no interval leaks), auto-scroll yields once the reader scrolls up, copy buttons report failure instead of claiming success.
- **CSS**: the five tabs fit at ≥1024px (fifth tab was clipped), stylesheet inlined via `?raw` so it ships only on this page (was a 15 KB extra `<link>` on ~900 unrelated pages), `--accent` renamed `--ash-accent` to stop shadowing the Hub's Tailwind token, brand tokens aliased, paused state freezes decorative animations, print rules, reduced-motion/no-JS window grows to show the whole transcript.
- **Content**: hero/tease no longer promise skills for Claude Desktop; scenes use real galaxy-mcp calls (`search_iwc_workflows` → `import_workflow_from_iwc` → `invoke_workflow`, `get_history_details`/`get_history_contents`), galaxy-mcp version `0.1.0`; indexable lede paragraph; git/Node.js prerequisites; `/tools/` blurb names the five shipped harnesses. The orphaned Cursor guide is removed while Cursor is held back (one-line revert in the exporter's `PAGES`).
- **CodeCopy** moved to BaseLayout so every prose route (events, platforms, bare) gets it and client-side navigation can't strand a page with the script but not the styles; announces "Copied"/"Copy failed" to assistive tech; no layout shift.
- **Tests** now assert the replay actually animates and auto-advances, pause semantics, the clipboard failure path, that reduced motion stays static, that guides keep their intro and link upstream, and that event pages get the copy button.

## Guides (same page)

Everything lives on the one page. Below the hero: a **Get a Galaxy API key** section, then **Set up your agent** with a tab strip (same five harnesses as the hero) and one panel per harness holding that harness's full guide (install, credentials, verify, skills, update, alternatives, troubleshooting). The hero tabs, guide tabs, chips, cards and the "Set up" button all select the same harness; the URL hash tracks it (`#guide-codex`), deep links open and scroll to the panel, and without JavaScript the panels render stacked. Auto-advance in the hero never changes the guide the reader has open.

The old per-harness URLs (`/tools/ai-agents/claude-code/` etc. and `/api-key/`) 301 to the matching anchor. The redirect generator now preserves `#fragment`s on targets.

Also links the section from `/tools/` next to the User-Defined Tools and Orbit entries.

The page is generated by the plugin repo's `scripts/export-hub.py` from its `docs/`, so the guides stay in sync with what actually ships; each panel links to its upstream source and the footer's edit link points there. The exporter's landing template must stay byte-identical to the hub page.

## Checks

- `npm run lint`, `format:check`, `content:lint` and a production `astro build` pass.
- New Playwright spec `astro/tests/ai-agents.spec.ts`: rendering, hero/guide tab sync, deep links, keyboard, replay animates + auto-advances, pause semantics, clipboard success and failure paths, reduced motion, phone-width overflow, copy buttons inside guide panels and on event pages. Existing article suites (content rendering, citations, print, static pages, FAIR, bare pages) pass.

The agentic stack page (`/tools/agentic-stack/`) is a separate PR stacked on this one: #4289.


content/tools/ai-agents/index.md
```diff
@@ -0,0 +1,825 @@
+---
+title: "Galaxy for AI Coding Agents"
+tease: "Connect Claude Code, Codex, Antigravity or Pi to your Galaxy server with curated Galaxy skills, or Claude Desktop with the Galaxy connection alone."
```
scottcain, line 3: I'm not exactly clear on what "Claude Desktop with the Galaxy connection alone"; I think maybe the tease could be rephrased to make it clearer.

scottcain: Additionally, I'm not thrilled with how the top of the page looks when rendered; the "One install: ..." part looks like a continuation of the tease (the text above it), but that makes the "intro text" cluttered and it doesn't read well. This might be fixed just by adding some white space, but I don't think there is an easy why to make that change in this file.

Image

content/tools/ai-agents/index.md
```diff
@@ -0,0 +1,825 @@
...
+subsites: [all]
+components: true
+autotoc: false
+skip_title_render: true
+full_bleed: true
+og_image: /images/galaxy-logos/galaxy_logo_25percent.png
+generated_from: https://github.com/galaxyproject/agentic-plugins/blob/main/scripts/export-hub.py
+---
+
+<AgentShells />
```
scottcain, line 13: The AgentShells thing is pretty neat.

scottcain (changes requested):
I added a few comments mostly for improving how it looks. I won't comment on the code itself (I'll leave that to @dannon ) but generally speaking, I like the content quite a bit

PR #4288, opened 2026-09-18
Introduce tag-specific feeds

wm75:
- For white-listed tags generate a combined news and events atom feed at /tags/[tag]/feed.atom analogous to the existing /tags/[tag] pages.
- Within such feeds distinguish news from events through <category> elements.
- List available tag-specific feeds in their own section of /feeds.
- Refactor common components of feed handling out into utils/feed.ts.
- Add tests for those utility functions.

This should be relevant for projects like esg or the upcoming esg4stars and others, for which adding an activity feed becomes as easy as adding the project tag to the whitelist.


wm75:
@dannon brc-analytics is not a very actively used tag, but is otherwise another good whitelist candidate if you're not producing a feed for the project elsewhere yet.

dannon (approved):
Pulled this down and verified locally -- builds cleanly and the feed utility refactoring is really nice.

Good call on flagging `brc-analytics` as a question, but let's hold off on that for now. There's really only one post tagged with it on the Hub so far, and most BRC updates/white papers live directly over on brc-analytics.org anyway. If we start channeling more project news through the Hub later on, tossing it into the whitelist is easy enough.

Looks great to merge from my end!

bgruening:
Should we move the tag-config to TAGS.yaml, so its in one place?

PR #4290, opened 2026-09-20
Icons and footers rework

wm75:


dannon (approved):
Much nicer to have it all in one spot, and the inline stuff gone. Freiburg footer matching looks great. Thanks Wolfgang!

PR #4302, opened 2026-09-22
Updating DH SIG page

Sch-Da:
With the link to the subdomain and newly added tools


content/community/sig/digital-humanities/index.md
```diff
@@ -48,13 +48,14 @@ Are you already familiar with Galaxy? Try one of our other tutorials or consider
   frameborder="0">
 </iframe>
 
+Scroll down to see all tutorials here.
```
bgruening, line 51: Is that really needed? And if we want to redirect users, should we not direct them to GTN?

Sch-Da: The scroll bar only appears on mouse over to show that there are more than 3 tutorials. So I thought I would make it more explicit that there are more tutorials visualised in the iframe. We can leave it out if you prefer.

